PR #3, opened 2018-09-12
STRIPES-553 Upgrade debug dependency

ttamj:


ttamj:
@MikeTaylor, I agree.  The Q3 release schedule occurred to me after issuing the PRs, so I went back to review the dates.  I will hold off on merging these.

PR #7, opened 2018-10-08
Nightmare coverage transfer branch

evaluk:
This has been tested in ui-users module as well as platform-core. This change requires action from developers that make unit/integration tests

1) they must not call end() in their tests
2) they must accept and use a nightmare parameter sent by the testing framework.  

known issue: Platforms do not produce coverage due to an unknown issue.  This is considered lower priority right now and may not be required.  Platforms run as expected with the coverage flag, there is no error involved.  It fails silently.


zburke:
This sounds great! Is there a branch in ui-users where we can see an example test that accepts nightmare as a param from the framework? Could you update the [README.md](/folio-org/stripes-testing/tree/master/README.md) file with that example?

evaluk:
Hi Zak, 
Yes there is a branch for ui-users with the changes that are needed: UIU-570.  I was unable to commit my updates to that branch even though I made the branch in the first place.  The current status is fine for education purposes but the code is old since I haven't revisited it in a while and it needed to be merged from master branch.  I should have that done today or tomorrow.

PR #10, opened 2019-01-28
Add the bigtest links suggested by Jeffrey Cherewaty on slack

julianladisch:


ttamj:
There is also this document in the `stripes` repo:
https://github.com/folio-org/stripes/blob/master/doc/bigtest.md

PR #12, opened 2019-02-15
add doc for running tests

zburke:
Write better documentation for running tests, particularly in the
context of folks who are not familiar with front-end code but who may
want to run the tests to check whether they pass based on changes to
back-end modules.

Refs [UITEST-56](https://issues.folio.org/browse/UITEST-56)

### DIFF
--- a/doc/nightmare.md
+++ b/doc/nightmare.md
@@ -5,7 +5,7 @@
 * [helpers](#helpers)
 * [namegen (helpers.js)](#namegen-helpersjs)
 * [openApp (helpers.js)](#openapp-helpersjs)
-* [Writing testable code]
+* [Writing testable code](#writing-testable-code)
   * [Provide unique identifiers on UI elements](#provide-unique-identifiers-on-ui-Elements)
 * [Writing robust tests](#writing-robust-tests)
   * [Avoid timers](#avoid-timers)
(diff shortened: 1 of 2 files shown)


doc/nightmare.md
```diff
@@ -1,5 +1,288 @@
-# Nightmare for FOLIO UI 
+# Nightmare for FOLIO UI
 
-To document
+* [Develop tests for a UI module](#develop-tests-for-a-ui-module)
+* [The test context object](#the-test-context-object)
+* [helpers](#helpers)
+* [namegen (helpers.js)](#namegen-helpersjs)
+* [openApp (helpers.js)](#openapp-helpersjs)
+* [Writing testable code]
```
ttamj: The link to this section is missing.  [md2toc](https://github.com/folio-org/okapi/blob/master/doc/md2toc) is your friend!

doc/nightmare.md
```diff
@@ -1,5 +1,288 @@
...
+interleaved with one another. Yes, really.
+
+
+### Extract text from the DOM using .evaluate()
+
+`.evaluate(fx, av1, ... avn)` accepts a function and a list of argument to pass
+to the function, and returns a Promise. You can use this feature to extract text
+from the DOM by passing a function that parses the DOM and returns a value. This
+example from platform-core's loan-renewal test gets the index of an anchor on
+the page, e.g. "<a href='some-link'>some text</a>", given "some text" and then
```
ttamj: You may need to escape the `<a>` tag here.  It generates as an actual link to nowhere when viewing the rendered markdown.

PR #19, opened 2019-03-26
grant-permission helper

zburke:


zburke:
Um, all we needed was to add the new function to the end of the file. Git, what happened here? How did you create a merge conflict out of that?

PR #38, opened 2019-08-15
UIIN-671 fix new-item integration helper

zburke:
# AN ODE TO NIGHTMARE
How do I loathe thee? Let me count they ways.
I loathe thee to the depth and breadth and height
Windows can reach, when render'd out of sight
Whether pixelated or by rays, trace'd.
I loathe thee to the level of every test's
Most quiet need: to PASS on every night
I loathe thee when I click an `<option>`, right?
Then click `submit` and lay my form to ReST
And find no `blur` event is ever fired
So validation never comes to pass
Three days, I lost, and nights, red eyes, I tired.
John called: "Set A, then B! It works! At last!
That blurs select and gets you out' the mire!"
'Twas he smashed the bug, but I'll kick your ass.

Alternatively, instead of interacting with a non-select element immediately after every select element in order to make sure its blur handlers fire, @JohnC-80 figured out that you can set the field's value and then fire focus/blur/change handlers on it inside an `evaluate` block. I tried that, and it works, but I've had Very Bad Experiences manipulating the DOM from within `evaluate` blocks; Nightmare sometimes gets lost in a Neverneverland of parallel execution. 

Finally, please note: you have to interact with a non-select element after _every_ call to `select()`. If you have two selects and a text field, you cannot traverse "select select text" you must traverse "select text select text". Does this make sense? No, not it does not. But does it work? Yes, yes it does. 

Fixes [UIIN-671](https://issues.folio.org/browse/UIIN-671)

### DIFF
--- a/helpers.js
+++ b/helpers.js
@@ -101,7 +101,7 @@ module.exports.getUsers = (nightmare, config, done) => function getu() {
 module.exports.createInventory = (nightmare, config, title, holdingsOnly) => {
   const ti = title || 'New test title';
   const id = new Date().valueOf();
-  const barcode = new Date().valueOf();
+  const barcode = `${new Date().valueOf()}${Math.floor(Math.random() * Math.floor(999999))}`;
   if (!holdingsOnly) {
     it('should create inventory record', (done) => {
       nightmare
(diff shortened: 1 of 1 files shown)


mkuklis:
I'm not sure why but this song comes to my mind after reviewing this PR:

https://www.youtube.com/watch?v=jhUkGIsKvn0

Amazing skills @JohnC-80 and @zburke!

PR #78, opened 2020-12-09
STCOM-785: interactors update - checkbox

jbolda:
This adds a new `CheckboxInteractor` which is coupled with https://github.com/folio-org/stripes-components/pull/1452.

/cc @JohnC-80


interactors/checkbox.js
```diff
@@ -0,0 +1,29 @@
...
+    label: (el) => el.textContent,
+    ariaLabel: (el) => el.querySelector('input')?.ariaLabel,
+    ariaInvalid: (el) => el.querySelector('input')?.getAttribute('aria-invalid') === 'true',
+    feedbackText: (el) => el.querySelector('[role=alert]').textContent,
+    hasWarning: (el) => !!el.className.match(/hasWarning/),
+    hasError: (el) => !!el.className.match(/hasError/),
+    disabled: (el) => el.disabled // we have cases where it is undefined and then what?
+  },
+  actions: {
+    ...CheckBox().specification.actions,
```
cowboyd, line 20: Are we sure this is what we want given that a checkbox wraps a single `input` element?

interactors/checkbox.js
```diff
@@ -0,0 +1,29 @@
...
+    focus: perform((el) => el.querySelector('label').focus()),
+    click: perform((el) => el.querySelector('label').click()),
+    clickInput: perform((el) => el.querySelector('input').click()),
```
cowboyd, line 23: could be re-written:

```js
clickInput: interactor => interactor.find(Checkbox()).click()
```

interactors/checkbox.js
jbolda: Good point. Removed.

interactors/checkbox.js
```diff
@@ -0,0 +1,38 @@
+import { CheckBox, createInteractor, perform } from '@bigtest/interactor';
+
+// The built-in CheckBox interactor expects an input element,
+// and anything added in the future would likely not work without modification.
+// Explicitly import known working functions.
+const { title, visible, focused } = CheckBox().specification.filters;
```
cowboyd, line 6: Does using this checkbox filter directly work? I would think that we'd want to see if the FOLIO checkbox component _contained_ the focus element:

```js
focus: el => el.contains(el.ownerDocument.activeElement)
```

jbolda: Good point, that is likely more all encompassing.

interactors/checkbox.js
jbolda: So the `input` element is actually not visible in this component. I think the `.find()` makes sense here, but I did have to use `Checkbox({ visible: false })`. That doesn't feel great, but I think it makes sense for this component.

JohnC-80 (approved):

PR #83, opened 2020-12-16
STCOM-785: interactors update: pane

jbolda:
# Motivation
Upgrade the Pane interactor to the new bigtest interactor syntax and make it available for consumption by any test suite. You can see this in action in folio-org/stripes-components#1462.


interactors/pane.js
```diff
@@ -0,0 +1,38 @@
...
+      for (let i = 0; i < panes.length; i++) {
+        if (el == panes[i]) {
+          return i;
+        }
+      }
+    },
+    focused
+  },
+  actions: {
+    dismiss: (interactor) => interactor.find(Button()).click(),
```
cowboyd, line 32: This will literally find _any_ button within the Pane. Is that specific enough?

doc/interactors.md
```diff
@@ -76,3 +79,46 @@ Accordion("Categories").exists()
...
+
+Pane("User Pane").is({ visible: true });
+```
+
+##### Locator
+
+Accordions are identified by their title. For example:
+
+```javascript
+Pane("User Pane").exists();
```
cowboyd, line 102: Panes do not usually have the word "Pane" in them. I'd use `Pane("Users")`

jbolda: I was thinking that too, but wondered if `"User"` was too vague.

interactors/pane.js
jbolda: Ah, sure enough. I will add a test that has multiple buttons as well to check this after I adjust it.

doc/interactors.md
cowboyd: I think in the context of someone using stripes, it's not. Our examples should reflect a plausible usage

cowboyd: I see your point though. Maybe `Pane("Search Results')` would be less ambiguous?

jbolda: Oh, that is good. Switching to that.

interactors/input.js
```diff
@@ -0,0 +1,19 @@
+import { createInteractor, perform, focused } from '@bigtest/interactor';
+import { isVisible } from 'element-is-visible';
+
```
cowboyd, line 3: Does the `TextField` interactor from `@bigtest/interactor` not work here?

jbolda: Oh sure, that makes sense.

PR #87, opened 2020-12-17
STCOM-785: interactors update: layer

jbolda:
# Motivation

Upgrade the Layer interactor to the new bigtest interactor syntax and make it available for consumption by any test suite. You can see this in action in folio-org/stripes-components#1464.


interactors/generic.js
```diff
@@ -0,0 +1,10 @@
+import { createInteractor, focused } from '@bigtest/interactor';
```
cowboyd, line 1: This should be called `Element`

jbolda: I was thinking in terms of the built-in elements, and I was a worried about confusion with the HTMLElement. Maybe not an issue?

cowboyd: Ah yeah, is there a global Element constructor? I think it’s fine to shadow it although it might screw up typescript.  🤔

jbolda: I switched it to `HTML`. The other contender was `Any`, but I think we were leaning towards `HTML` since it is a little more descriptive.

doc/interactors.md
```diff
@@ -40,6 +40,7 @@ include the DOM under test such as Selenium, or Nightmare.
 - [`KeyValue`](#keyvalue)
 - [`Select`](#select)
 - [`Tooltip`](#tooltip)
+- [`Layer`](#layer)
```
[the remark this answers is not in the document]
jbolda: Oops, that poor timing, just merged. Can do it in a follow-up PR.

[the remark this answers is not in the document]
jbolda: Or rather, I will rebase #89 and alphabetize there.

PR #90, opened 2020-12-22
STCOM-785: interactors update: text-field

jgnieuwhof:
## Motivation
Upgrade the `TextField` interactor to the new bigtest interactor syntax, and make it available for consumption by the test suite.

## Approach
Added a `TextField` interactor, with filters and actions that cover the common ways of interacting with a text field.

See this in action here: https://github.com/folio-org/stripes-components/pull/1468

### DIFF
--- a/interactors/index.js
+++ b/interactors/index.js
@@ -10,5 +10,4 @@ export { default as Select } from './select';
 export { default as Label } from './label';
 export { default as Button } from './button';
 export { default as Pane } from './pane';
-
-export { TextField } from '@bigtest/interactor';
+export { default as TextField } from './text-field';
--- a/interactors/text-field.js
+++ b/interactors/text-field.js
@@ -1,0 +1,32 @@
+import { createInteractor, perform, TextField } from '@bigtest/interactor';
+
+import IconButton from './icon-button';
+
+const label = (el) => {
+  const labelText = el.querySelector('label');
+  return labelText ? labelText.innerText : undefined;
+};
+
+export default createInteractor('text field')({
+  selector: 'div[class^=textField-]',
+  locator: label,
+  filters: {
+    id: (el) => el.querySelector('input').id,
+    label,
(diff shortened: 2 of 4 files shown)


interactors/text-field.js
```diff
@@ -0,0 +1,35 @@
...
+export default createInteractor('text field')({
+  selector: 'div[class^=textField-]',
+  locator: label,
+  filters: {
+    id: (el) => el.querySelector('input').id,
+    label,
+    type: (el) => el.querySelector('input').type,
+    value: (el) => el.querySelector('input').value,
+    focused: (el) => el.querySelector('input').contains(el.ownerDocument.activeElement),
+    readOnly: (el) => el.querySelector('input').hasAttribute('readOnly'),
```
jbolda, line 19: /cc @elrickvm, maybe this is better than the double bang we were doing last night.

interactors/index.js
```diff
@@ -10,5 +10,4 @@ export { default as Select } from './select';
 export { default as Label } from './label';
 export { default as Button } from './button';
 export { default as Pane } from './pane';
-
-export { TextField } from '@bigtest/interactor';
```
jbolda, line 14: Do we need to adjust the tests that relied on this lower level version of `TextField`?

jgnieuwhof: I don't think so 🤔. I did some grepping, couldn't find this in-use, and the test suite is passing locally.

jbolda: Looks like I might have added it, switched to a different direction and then never removed it? Appears that we don't need to do anything with it.

jbolda (approved):

interactors/text-field.js
elrickvm: nice!

elrickvm (approved):

cowboyd (approved):

PR #91, opened 2020-12-23
STCOM-785: interactors update: textarea

jgnieuwhof:
## Motivation
Upgrade the `TextArea` interactor to the new bigtest interactor syntax, and make it available for consumption by the test suite.

## Approach
Added a `TextArea` interactor, with filters and actions that cover the common ways of interacting with a text field.

See this in action here: https://github.com/folio-org/stripes-components/pull/1470

### DIFF
--- a/interactors/textarea.js
+++ b/interactors/textarea.js
@@ -1,0 +1,27 @@
+import { createInteractor, perform } from '@bigtest/interactor';
+
+const label = (el) => {
+  const labelText = el.querySelector('label');
+  return labelText ? labelText.innerText : undefined;
+};
+
+export default createInteractor('text area')({
+  selector: 'div[class^=textArea-]',
+  locator: label,
+  filters: {
+    id: el => el.querySelector('textarea').getAttribute('id'),
+    label,
+    value: el => el.querySelector('textarea').value,
+    warning: el => el.querySelector('[class^=feedbackWarning-]').textContent,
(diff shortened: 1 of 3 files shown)


interactors/textarea.js
```diff
@@ -0,0 +1,27 @@
...
+    warning: el => el.querySelector('[class^=feedbackWarning-]').textContent,
+    error: el => el.querySelector('[class^=feedbackError-]').textContent,
+    valid: el => el.querySelector('textarea').getAttribute('aria-invalid') !== 'true'
+  },
+  actions: {
+    blur: perform(el => el.querySelector('textarea').blur()),
+    fillIn: async (interactor, value) => {
+      await interactor.focus();
+      await interactor.perform(el => { el.querySelector('textarea').value = value; });
+    },
```
cowboyd, line 24: This seems like something that would be good to delegate to the native `TextArea` component:

```js
fillIn: (interactor, value) => interactor.find(TextArea()).fillIn(value)
```

cowboyd (commented):
Looking good. Only one suggestion on the delegation of `fillIn` duties. The reason that delegation is a good thing is that the native `TextArea` component actually does all the subtle things like firing focus, keydown, keyup, input, and change events.

Also,  Let's insert the docs in alphabetic order.

cowboyd (approved):

PR #92, opened 2020-12-23
STCOM-785: docs for new interactors for button and icon button

jgnieuwhof:
The original implementation was added here: https://github.com/folio-org/stripes-testing/pull/81

Now that we have a documentation format it's time to backfill some docs.


cowboyd (approved):
📃🖋😍

PR #93, opened 2020-12-28
STCOM-785: interactors update: searchField

elrickvm:
### Motivation

Upgrade the `SearchField` interactor to the new bigtest interactor syntax. Make it available for usage in the test suite.

### Approach

Added a `SearchField` interactor, with filters and actions for interacting with a search-field.

See this in action here:

### DIFF
--- a/doc/interactors.md
+++ b/doc/interactors.md
@@ -298,7 +298,13 @@ import { SearchField } from '@folio/stripes-testing';
 
 SearchField().has({ id: 'searchFieldTest'});
 ```
+##### Locator
 
+A SearchField is located by ___. For example:
+
+```javascript
+SearchField("Label").exists();
+```
 ##### Actions
 
 - `selectIndex(string)`: selects the searchable index to perform this search on. For example "users" or "locations"
--- a/doc/interactors.md
+++ b/doc/interactors.md
@@ -41,6 +41,7 @@ include the DOM under test such as Selenium, or Nightmare.
 - [`Layer`](#layer)
 - [`Pane`](#pane)
 - [`RadioButton`](#radiobutton)
+- [`SearchField`](#searchField)
 - [`Select`](#select)
 - [`TextField`](#textfield)
 - [`Tooltip`](#tooltip)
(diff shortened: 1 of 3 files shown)


doc/interactors.md
```diff
@@ -286,6 +287,35 @@ RadioButton("Label").exists();
...
+
+- `id`: _string_ = the DOM element id of the contained `select`
+  element.  Note that this _not the id of the React component_. The `id`
+  filter is provided for debugging, but should not generally be used
+  in tests since it is not a user-facing value.
+- `placeholder`: _string_ = The placeholder text of the contained `input` element.
+- `value`: _string_ = the `value` property of the contained `input` element.
+- `readOnly`: _boolean_ = `true` if the contained `input` is read-only.
+- `isDisabledIndex`: _boolean_ = `true` if the contained `select` is disabled.
+- `isDisabled`: _boolean_ = `true` if the contained `input` is disabled.
```
cowboyd, line 317: let's just have a single filter `disabled: boolean` since that is really the only user facing property there is on the component. When testing that the `select` is disabled, that only happens when the search field is loading, and so needing to make assertions on this only needs to happen in the context of the stripes-components test suite and we can use `interactor.assert()` to handle this one-off.

doc/interactors.md
```diff
@@ -286,6 +287,35 @@ RadioButton("Label").exists();
...
+
+``` javascript
+import { SearchField } from '@folio/stripes-testing';
+
+SearchField().has({ id: 'searchFieldTest'});
+```
+
+##### Actions
+
+- `selectIndex(string)`: selects the option matching option name
```
cowboyd, line 304:
```suggestion
- `selectIndex(string)`: selects the searchable index to perform this search on. For example "users" or "locations"
```

doc/interactors.md
```diff
@@ -286,6 +287,35 @@ RadioButton("Label").exists();
 - `hasWarning`: _boolean_ = `true` if the radio button has a warning [class]
 - `hasError`: _boolean_ = `true` if the radio button has an error [class]
 
+#### SearchField
+
+Stripes SearchField component composes basic `<select>` and `<input>` elements
```
cowboyd, line 292: I'd suggest less how, more what.

```suggestion
Stripes SearchField component for initiating queries
```

cowboyd (commented):
Looking good! Can we add a locator section to the docs?

elrickvm:
@cowboyd I was not sure how to use the locator in this context / interactor but I could add one.

interactors/search-field.js
```diff
@@ -0,0 +1,22 @@
...
+
+export default createInteractor('search field')({
+  selector: '[class^=searchField]',
+  locator: label,
+  filters: {
+    id: el => el.querySelector('input').getAttribute('id'),
+    readOnly: el => el.querySelector('input').hasAttribute('readOnly'),
+    value: el => el.querySelector('input').value,
+    placeholder: el => el.querySelector('input').placeholder,
+    isDisabled: el => el.querySelector('select').disabled,
```
cowboyd, line 16: Convention is to not put `is` in front of filters, that way it reads nicely with the `is()` assertion: E.g.

```js
SearchField("users").is({ disabled: false })
```

```suggestion
    disabled: el => el.querySelector('select').disabled,
```

cowboyd (commented):
Looks, great, although one suggestion about the naming of filters

doc/interactors.md
```diff
@@ -286,6 +287,40 @@ RadioButton("Label").exists();
...
+##### Filters
+
+- `id`: _string_ = the DOM element id of the contained `select`
+  element.  Note that this _not the id of the React component_. The `id`
+  filter is provided for debugging, but should not generally be used
+  in tests since it is not a user-facing value.
+- `placeholder`: _string_ = The placeholder text of the contained `input` element.
+- `value`: _string_ = the `value` property of the contained `input` element.
+- `readOnly`: _boolean_ = `true` if the contained `input` is read-only.
+- `isDisabled`: _boolean_ = `true` if the contained `input` is disabled.
```
cowboyd, line 322:
```suggestion
- `disabled`: _boolean_ = `true` if the contained `input` is disabled.
```

PR #94, opened 2020-12-29
STCOM-785: interactors update: dropdown

jgnieuwhof:
## Motivation
Upgrade the `Dropdown` interactor to the new bigtest interactor syntax, and make it available for consumption by the test suite.

## Approach
Added a `Dropdown` interactor, with filters and actions that cover the common ways of interacting with a text field.

See this in action here: https://github.com/folio-org/stripes-components/pull/1473

### DIFF
--- a/interactors/dropdown.js
+++ b/interactors/dropdown.js
@@ -16,19 +16,32 @@ const DropdownMenu = createInteractor('dropdown menu')({
   }
 });
 
+const open = el => el.querySelector('[aria-haspopup]').getAttribute('aria-expanded') === 'true';
+
+const control = (shouldOpen = true) => async interactor => {
+  let isOpen;
+  await interactor.perform(el => { isOpen = open(el); });
+  if (isOpen !== shouldOpen) {
+    await interactor.toggle();
+  }
+};
+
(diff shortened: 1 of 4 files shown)


interactors/dropdown.js
```diff
@@ -0,0 +1,50 @@
...
+
+const open = el => el.querySelector('[aria-haspopup]').getAttribute('aria-expanded') === 'true';
+
+// todo: 'visible' should ensure the DropdownMenu is visible as well,
+// once filters support interactor composition
+const visible = el => [el, el.querySelector(['[aria-haspopup]'])].every(isVisible);
+
+const control = (shouldOpen = true) => async interactor => {
+  let isOpen;
+  await interactor.perform(el => { isOpen = open(el); });
```
jgnieuwhof, line 29: @cowboyd - this is an escape hatch for checking the value of a filter, without an assertion.

is there an easier way to do this?

cowboyd, line 29: We should probably allow `perform` to return a value. Doesn't affect this PR, but wanted to note that.

cowboyd (approved):

PR #95, opened 2021-01-05
STCOM-785: interactors update: MultiColumnList

jgnieuwhof:
### Motivation

Upgrade the `MultiColumnList` interactor to the new bigtest interactor syntax. Make it available for usage in the test suite.

### Approach

Added a `MultiColumnList` interactor, with filters and actions for interacting with a search-field.

See this in action here: https://github.com/folio-org/stripes-components/pull/1481


doc/interactors.md
```diff
@@ -182,7 +183,7 @@ The checkbox element
 
 ##### Synopsis
 
-```javascript
+````javascript
```
cowboyd, line 186: I think we only want three ` characters

doc/interactors.md
```diff
@@ -257,8 +258,8 @@ The IconButton element
 
 ##### Synopsis
 
-``` javascript
-import { IconButton } from '@folio/stripes-testing';
+```javascript
+import { IconButton } from "@folio/stripes-testing";
```
cowboyd, line 262: What is the convention here in FOLIO land? we should stick with that.

doc/interactors.md
```diff
@@ -353,6 +354,82 @@ None.
...
+Alternatively when only one is rendered on the page, you will likely prefer to omit the locator.
+
+```javascript
+MultiColumnList().exists();
+```
+
+###### Actions
+
+- `clickHeader`: clicks on the overall header element
+- `scrollBy`: _int_ = scrolls the container by the specified pixels entered as an integer, will trigger the related scroll events
```
cowboyd, line 387: We don't use `=` in the action by default
```suggestion
- `scrollBy`: _int_ = scrolls the container by the specified pixels entered as an integer, will trigger the related scroll events
```

cowboyd (approved):

doc/interactors.md
jbolda: Autoformat goofing things up compensating for another errant backtick. Also thought I had that off, but I turned it off in the wrong settings file.

doc/interactors.md
jbolda: Single quotes appears to be the way.

PR #97, opened 2021-01-08
STCOM-785: interactors update: Datepicker

jbolda:
## Motivation
Upgrade the Datepicker and Calendar interactor to the new bigtest interactor syntax. Make it available for usage in the test suite.

## Approach
Added a Datepicker and Calendar interactor, with filters and actions for interacting with both the datepicker input field and the calendar widget.

The related `stripes-components` PR will not be opened until this one has been merged in. The related code can be seen [here](https://github.com/jbolda/stripes-components/commit/98583a519e01b65f59267426e828e029ebf62e21) until it is opened.

## Outstanding
Note the tests that are currently skipped. Overall we are missing some pieces to test the keyboard functionality. This will likely fall into the keyboard interactor and not be in the datepicker interactor, but still makes sense to include it in the scope of this PR. 

It also makes sense to include `CalendarDays` more fully within the `Calendar` interactor and expose it as a public API, but need to work out what that will look like.

### DIFF
--- a/interactors/keyboard.js
+++ b/interactors/keyboard.js
@@ -1,6 +1,6 @@
 /* eslint-disable no-use-before-define */
 
-import { createInteractor, perform } from '@bigtest/interactor';
+import { createInteractor, perform, Select } from '@bigtest/interactor';
 
 export default createInteractor('keyboard (requires window to have focus)')({
   selector: ':focus',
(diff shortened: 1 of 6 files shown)


cowboyd (approved):

cowboyd (commented):
Although I forgot to mention the documentation. Although it all looks good to me!

interactors/keyboard.js
```diff
@@ -9,6 +9,38 @@ export default createInteractor('keyboard (requires window to have focus)')({
     arrowDown: press('ArrowDown'),
     arrowLeft: press('ArrowLeft'),
     arrowRight: press('ArrowRight'),
+    // using an underscore to signify that this is a temporary workaround
+    // we want to use the default browser implementation, but it doesn't
+    // trigger the built-in browser functions so we will need a more
+    // fleshed out approach to this
+    _nextOption: perform(async (el) => {
```
cowboyd, line 16: If we ask folks to use it, then it's a part of the public API and so we should just go ahead and own it, and then deprecate it when we have a better solution.

jbolda: That is fair. Do you then think I should split it into a `<select>` and a `<input type="number">` version instead of the overloaded version I have here?

PR #105, opened 2021-04-21
FOLIO-2946 initialize cypress tests

NikitaSedyx:
https://issues.folio.org/browse/FOLIO-2946


.gitignore
```diff
@@ -2,3 +2,7 @@ node_modules
 yarn.lock
 .yarnclean
 yarn-error.log
+
+.bigtest
```
NikitaSedyx, line 6: TODO: remove not required bigtest related items

cypress/support/bigtest.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+});
+
+Cypress.Commands.add('expect', (interaction) => {
+  if (Array.isArray(interaction)) {
+    interaction.map(i => interact(i, 'assertion'));
+  } else {
+    interact(interaction, 'assertion');
+  }
+});
```
cowboyd, line 32: This is the integration we provide in `@bigtest/cypress`

How can we improve that package so that you don't have to copy and paste that code?

NikitaSedyx: I think it's better to expose just your commands like other do https://docs.cypress.io/plugins/directory#Custom%20Commands

aliaksei-chumakou (approved):

cypress/support/bigtest.js
cowboyd: We followed the template of plugins like https://github.com/RcKeller/cypress-redux and https://github.com/component-driven/cypress-axe

So if you add 
```js
import `@bigtest/cypress`;
```

 it should register the `do` and `expect` commands for you.

NikitaSedyx: nice, I'll try to avoid this copy paste from original repo, makes sense

NikitaSedyx: yep, thanks for pointing  this

NikitaSedyx:
hi @zburke @JohnC-80 could you please take a look?

cypress/support/checkin.js
```diff
@@ -0,0 +1,20 @@
...
+  Checkbox,
+  MultiColumnList,
+  TextField,
+} from '../../interactors';
+
+Cypress.Commands.add('checkInItem', barcode => {
+  cy.do([
+    TextField({ id: 'input-item-barcode' }).fillIn(barcode),
+    Button('Enter').click(),
+    Button({ text: 'Check in', button: true }).click(),
```
ViktorSoroka07, line 12: Please clarify why for Button the prop `button` should be passed?

ViktorSoroka07:
`cypress/.DS_Store` file should be removed, right?

cypress/support/stripes.js
```diff
@@ -0,0 +1,26 @@
...
+  body,
+}) => {
+  const queryString = (new URLSearchParams({
+    ...DEFAULT_SEARCH_PARAMS,
+    ...searchParams,
+  })).toString();
+
+  cy.request({
+    method,
+    url: `${Cypress.env('OKAPI_HOST')}/${path}?${queryString}`,
```
ViktorSoroka07, line 19: From where OKAPI_HOST variable is going to be taken?

cypress/integration/checkout/checkout.basic.spec.js
```diff
@@ -0,0 +1,65 @@
...
+  });
+
+  after(() => {
+    cy.logout();
+  });
+
+  it('Basic flow', function () {
+    cy.visit('/checkout');
+
+    cy.checkOutItem(Cypress.env('users')[0].barcode, ITEM_BARCODE);
```
ViktorSoroka07, line 61: What is the difference between this approach? Wondering whether one approach can be used.

https://github.com/folio-org/stripes-testing/pull/105/files#diff-ba340d01f94ccbd02897d172ef5abadd0dbd2b2c2e4e22cb44c9b43c5997d26eR53-R57

Overall I see the setup for check-in is almost the same, only the [query differs](https://github.com/folio-org/stripes-testing/pull/105/files#diff-ba340d01f94ccbd02897d172ef5abadd0dbd2b2c2e4e22cb44c9b43c5997d26eR25) and a [title](https://github.com/folio-org/stripes-testing/pull/105/files#diff-ba340d01f94ccbd02897d172ef5abadd0dbd2b2c2e4e22cb44c9b43c5997d26eR33) for the item. Wondering whether it makes sense to create a helper and use it in two places.

NikitaSedyx:
> cypress/.DS_Store file should be removed, right?

hm, I made a commit to remove this, interesting

cypress/support/stripes.js
NikitaSedyx: https://wiki.folio.org/pages/viewpage.action?pageId=62555156
configuration section

cypress/integration/checkout/checkout.basic.spec.js
NikitaSedyx: > What is the difference between this approach? Wondering whether one approach can be used.

here we test checkout, in check in test we use API to create item with `Check out` status to avoid UI interactions - API is much faster that UI

> Overall I see the setup for check-in is almost the same
yeah, this PR to initialize work and provide a working example

cypress/support/checkin.js
cowboyd: Is there a "Check in" link as well?

NikitaSedyx: yep, menu item in navigation is `Check in`, and button on the screen `Check in`

ViktorSoroka07 (approved):

cypress/support/checkin.js
zburke: See, this is the reason I get nervous about using anything other than element-ids as selectors. I mean, I'm glad
```
Button({ button: true })
```
works, but without comments explaining why it's there (@NikitaSedyx's comment 1) and what purpose it's serving (comment 2) this line is very confusing to somebody just reading the test. 

Maybe it's just the nature of UI testing that there is tension between identifiers that are unique in the DOM but hidden from users (and therefore dangerous because the test can interact with the DOM in a way the user can't) and selectors that must specify DOM attributes that are hidden from users (and therefore brittle because they may rely on DOM attributes the user never sees and that the developer did not likely intentionally expose as an API surface and therefore are likely to change with notice). 🤷 

Anyway, I think these comments should be copied into the code 🙂

zburke (approved):

cypress/support/checkin.js
cowboyd: You could also scope the interactor which would make more sense. Like say you new that the button was in a certain `Pane`, you could say:

`Pane("Main").find(Button("Check in"))`

which would disambiguate it from the check in button in the menu in a natural feeling way.

NikitaSedyx: nice, in our case it's modal, but idea is the same, thanks!

cypress/support/checkin.js
```diff
@@ -9,7 +10,7 @@ Cypress.Commands.add('checkInItem', barcode => {
   cy.do([
     TextField({ id: 'input-item-barcode' }).fillIn(barcode),
     Button('Enter').click(),
-    Button({ text: 'Check in', button: true }).click(),
+    Modal('Confirm multipiece check in').find(Button('Check in')).click(),
```
cowboyd, line 13: Just a heads up, if you want to make the Modal match a little more flexible, you can use the `including()` matcher:

```js
Modal(including('Confirm')).find(Button('Check in')).click()
```

For more on matchers https://frontside.com/bigtest/docs/interactors/matchers#writing-flexible-assertions-using-matchers

PR #113, opened 2021-07-20
Stripes components refresh

JohnC-80:
This PR refreshes the `bigtest` dependency and re-adds `@bigtest/interactor` to dependencies.
`stripes-testing` is typically a devDependency of modules making use of it, but it needs `@bigtest/interactor` as a dependency in order for its required version to be installed - particularly when a consuming module already has tests that use the old version (>0.10.0.. most use 0.7.2) of `@bigtest/interactor`

Additionally, minor updates to interactors.
The default HTML interactor from `@bigtest/interactor` has `{visible: true}` as a default filter. This ends up failing to match if an extended interactor doesn't override the `visible` filter.


cowboyd:
Shouldn't the default still be visible? In other words, this will cause interactors to match and interact with elements that may not be visible on screen.

JohnC-80:
@cowboyd Thanks for the comment! I can understand why the default is `visible: true` for interacting with applications/components as users do. That's totally okay and understandable - but maybe some documentation is necessary (if it doesn't already explain it) because it did catch me a little off guard.

Say you're testing functionality of dropdowns or accordions and you want to test that an element is not visible - that the menu/accordion is closed. If you extend HTML for your purposes, but you don't override the `visible` filter, attempting to use `ddMenu.is({ visible: false })` will give you the error message that it couldn't find something matching `visible = true` (via the nicely presented console tables.)  - and it'll vex you for a just little bit because you swear `visible: false` is what you asked for - until you check out the source and see the way that the default is set up.

Maybe the overrides should exist in test-exclusive interactors - I'd be okay with that as it's component-specific functionality - but maybe someone does have app-level functionality that closes (or hides) accordions that contain no data...

Or maybe I'm just totally missing something on how the filters are inherited and there's generally a better way to do this.

zburke:
I [hit this `visible: true` surprise too](https://discord.com/channels/700803887132704931/783795154967330857/817154424819417108).

wKich:
I agree the visibility filter is confusing in some cases.

cowboyd:
This is the part that's throwing me for the loop:

> If you extend HTML for your purposes, but you don't override the visible filter, attempting to use ddMenu.is({ visible: false }) will give you the error message that it couldn't find something matching visible = true (via the nicely presented console tables.) 

In other words, you are making the assertion that visible is false, but you cannot find the element to make the assertion because visibility is false.

The workaround here is that you would use the `Menu({ visible: false }).exists()` form, but that doesn't read that well, and it leaves in place the gotcha.

Is there no way to do something like if you're asserting on filter to remove its default value from the lookup?

cowboyd:
In other words, we have a set of filters that we use for resolving the element, and a set of filters upon which we make assertions

So when we say `Button('New')`, what we're really doing is saying behind the scenes `Button('New', { visible: true, disabled: false })` and when we say `Button('New').is({ visible: false })` what this (currently) says behind the scenes is `Button('New', { visible: true, disabled: false }).is({ visible: false })` which is obviously an incoherent assertion.

The question I'm asking then is this: "Would it not be a better API to make it impossible to write such incoherent assertions?" In other words, because we detect a value for the `visible` filter in the assertion, we then remove the filter value from the lookup.

If we did that, then `Button('New').is({ visible: false })` would be shortand for `Button('New', {disabled: false}).is({ visible: true })`

interactors/multi-column-list.js
```diff
@@ -13,7 +22,9 @@ export const MultiColumnListCell = HTML.extend('multi column list cell')
     column: (el) => el.textContent,
     columnIndex: childIndex,
     selected: (el) => !!el.parentElement.className.match(/mclSelected/),
-    measured: (el) => el.style && el.style.width !== ''
+    measured: (el) => el.style && el.style.width !== '',
+    visible: (el) => isVisible(el),
+    clientTop: el => el.getBoundingClientRect().top,
```
cowboyd, line 27: It's always an art with interactors whether to include very render-specific properties. In fact, we've found a fundamental tension in other projects between interactors filters useful to the author of the component-library and design system interactor filters that are useful to application developers. It seems like the former is a superset of the latter. More simply, it's a question whether in any stripes apps, that you would need the `clientTop` property, but it is very useful in testing `stripes-components` itself. If an app developer would never use it, then its presence in the API can be confusing.

One way to relieve this tension is to make an `element` filter in your base interactor that exposes the underlying element. This isn't that useful by itself since you rarely know the element ahead of time (after all that's kinda what interactors are for). However, this does allow you use [matchers][matchers] implemented locally within your test suite against your element which has two implications:

1. You don't need to add render-specific filters that nobody will actually use outside of the component test suite.
2. You can use the matcher on _all_ of your interactors, not just the one that has the filter (in this case multi column list cell)

For example, let's say you had a matcher called `withClientTop`:

```js
export function withClientTop(value) {
  return {
    match(actual) {
      return actual.getBoundingClientRect().top === value;
    },
    description() {
      return `clientTop === ${value}`;
    }
  };
};
```

Now, you can use this anywhere:

```js
await MultiColumnListCell().has({ element: withClientTop(100) });
await Accordion().find(Button()).has({ element: withClientTop(55) });
```

If that's an option you'd want to take, then you could do it pretty easily by exporting your own `HTML` interactor from the index:

```js
import { HTML as BaseHTML } from '@bigtest/interactor';

export const HTML = BaseHTML
  .filters({
    element: element => element
  });
```

and having all the internal components use that as their base (consumers of stripes-testing would get this transparently since they already import `HTML` from folio, not bigtest)

[matchers]: https://frontside.com/bigtest/docs/interactors/matchers

cowboyd (commented):
This PR highlights what I think is a very valid issue with how we handle default values are handled for identification and assertion and what happens when those two overlap. I'm going to propose that we change the behavior in bigtest core, but in the meantime, I think it's totally fine to use the workaround you have since the current behavior is strange to say the least.

In fact, if you opt for the `HTML` override, then you could make this the workaround for all interactors in order to have consistent behavior. I'll leave that up to you though.

cowboyd:
> @cowboyd
> 
> > If we did that, then Button('New').is({ visible: false }) would be shortand for Button('New', {disabled: false}).is({ visible: true })
> 
> I think you wanted to say 'visible: false', didn't it?

I did 😄 

will update comment....

JohnC-80:
Hey @cowboyd, @wKich .. thanks for the feedback!  Interactors at the library level vs interactors at the app level has always been a bit of a struggle for me.

TIL: Custom matchers. Woah. Something I've only just seen so far but haven't gotten in to understand / customize so much.

Some good news: bringing in this branch actually has the test suite running with these new interactors over on https://github.com/folio-org/stripes-components/pull/1596 !
The bad news - weird error happening in that textfield test... will reach out/tag y'all on the review...

zburke (approved):
👏 👏 👏 🎉  Nice work here, @JohnC-80, as always!

cowboyd (approved):
Looks good to me!

PR #114, opened 2021-07-26
FAT-901 Add Remote Storage Configurations CRUD happy path

axelhunn:
https://issues.folio.org/browse/FAT-901

### DIFF
--- a/cypress/integration/remote-storage/configurations/crud.spec.js
+++ b/cypress/integration/remote-storage/configurations/crud.spec.js
@@ -3,7 +3,6 @@ import { Link } from '@bigtest/interactor';
 import {
   Button,
   KeyValue,
-  Layer,
   Modal,
   MultiColumnList,
   MultiColumnListCell,
@@ -14,7 +13,7 @@ import {
 
 bigtestGlobals.defaultInteractorTimeout = 10000;
 
-describe('Configurations CRUD', () => {
+describe('making CRUD', () => {
   before(() => {
     cy.visit('/');
 
@@ -36,7 +35,7 @@ describe('Configurations CRUD', () => {
     // Create
     cy.do(Button('+ New').click())
       .expect([
-        Layer('Create configuration').exists(),
+        Pane('Create configuration').exists(),
         TextField('Remote storage name\n*').is({ focused: true, value: '' }),
         Select('Provider name\n*').has({ value: '' }),
         Button('Cancel').exists(),
@@ -50,14 +49,14 @@ describe('Configurations CRUD', () => {
       ])
       .expect([
         Modal('Create configuration').exists(),
-        Button('Cancel').exists(),
-        Button('Save').exists(),
+        Modal('Create configuration').find(Button('Cancel')).exists(),
+        Modal('Create configuration').find(Button('Save')).exists(),
       ])
 
       .do(Button('Save').click())
       .expect([
         Modal('Create configuration').absent(),
-        Layer('Create configuration').absent(),
+        Pane('Create configuration').absent(),
         MultiColumnList('storages-list').exists(),
         MultiColumnListCell(NAME).exists(),
       ]);
@@ -66,7 +65,7 @@ describe('Configurations CRUD', () => {
     cy.do(MultiColumnListCell(NAME).click())
       .expect([
         Pane(NAME).exists(),
-        Button('Actions').exists(),
+        Pane(NAME).find(Button('Actions')).exists(),
         KeyValue('Remote storage name').has({ value: NAME }),
         KeyValue('Provider name').has({ value: PROVIDER }),
       ]);
@@ -77,7 +76,7 @@ describe('Configurations CRUD', () => {
 
       .do(Button('Edit').click())
       .expect([
-        Layer(`Edit ${NAME}`).exists(),
+        Pane(`Edit ${NAME}`).exists(),
         TextField('Remote storage name\n*').is({ focused: true, value: NAME }),
         Select('Provider name\n*').has({ value: PROVIDER_CODE }),
         Button('Cancel').exists(),
@@ -90,14 +89,14 @@ describe('Configurations CRUD', () => {
       ])
       .expect([
         Modal(`Edit ${NAME}`).exists(),
-        Button('Cancel').exists(),
-        Button('Save').exists(),
+        Modal(`Edit ${NAME}`).find(Button('Cancel')).exists(),
+        Modal(`Edit ${NAME}`).find(Button('Save')).exists(),
       ])
 
       .do(Button('Save').click())
       .expect([
         Modal(`Edit ${NAME}`).absent(),
-        Layer(`Edit ${NAME}`).absent(),
+        Pane(`Edit ${NAME}`).absent(),
         MultiColumnListCell(NAME).absent(),
         MultiColumnListCell(NAME_EDITED).exists(),
       ])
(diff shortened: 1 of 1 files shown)


cypress/integration/remote-storage/configurations/crud.spec.js
```diff
@@ -0,0 +1,127 @@
...
+        Button('Save & close').exists(),
+      ])
+
+      .do([
+        TextField('Remote storage name\n*').fillIn(NAME),
+        Select('Provider name\n*').choose(PROVIDER),
+        Button('Save & close').click(),
+      ])
+      .expect([
+        Modal('Create configuration').exists(),
```
cowboyd, line 51: This is redundant since you're asserting that both the save and cancel buttons exist within this modal, you will get an error message specifically highlighting the absence of the 'create configuration' modal if it does not exist. (it won't even look for the save and cancel buttons because it will short circuit)

Since the next block of actions uses the save button, you don't need to verify that it exists, or wait to proceed until it exists. The act of using it will wait wait until it exists, is visible, and is enabled, and if it is not all of those things you will get a precise error message detailing why. Also, if it cannot be found it will look for other buttons on the page and present them as possible alternatives.

cypress/integration/remote-storage/configurations/crud.spec.js
```diff
@@ -0,0 +1,127 @@
...
+    const PROVIDER_CODE = 'DEMATIC_EMS';
+
+    // Create
+    cy.do(Button('+ New').click())
+      .expect([
+        Pane('Create configuration').exists(),
+        TextField('Remote storage name\n*').is({ focused: true, value: '' }),
+        Select('Provider name\n*').has({ value: '' }),
+        Button('Cancel').exists(),
+        Button('Save & close').exists(),
```
cowboyd, line 42: This is redundant since you'll be using this button later. In most cases, you do not need `exists()` assertions.

cypress/integration/remote-storage/configurations/crud.spec.js
```diff
@@ -0,0 +1,127 @@
...
+
+      .do([
+        TextField('Remote storage name\n*').fillIn(NAME),
+        Select('Provider name\n*').choose(PROVIDER),
+        Button('Save & close').click(),
+      ])
+      .expect([
+        Modal('Create configuration').exists(),
+        Modal('Create configuration').find(Button('Cancel')).exists(),
+        Modal('Create configuration').find(Button('Save')).exists(),
```
cowboyd, line 53: As mentioned in the previous comment, this line is redundant

cypress/integration/remote-storage/configurations/crud.spec.js
```diff
@@ -0,0 +1,127 @@
...
+    cy.do(Button('Actions').click())
+      .expect(Button('Edit').exists())
+
+      .do(Button('Edit').click())
+      .expect([
+        Pane(`Edit ${NAME}`).exists(),
+        TextField('Remote storage name\n*').is({ focused: true, value: NAME }),
+        Select('Provider name\n*').has({ value: PROVIDER_CODE }),
+        Button('Cancel').exists(),
+        Button('Save & close').exists(),
```
cowboyd, line 83: This is redundant since the test uses the save & close button

cypress/integration/remote-storage/configurations/crud.spec.js
```diff
@@ -0,0 +1,127 @@
...
+      ])
+
+      .do([
+        TextField('Remote storage name\n*').fillIn(NAME_EDITED),
+        Button('Save & close').click(),
+      ])
+      .expect([
+        Modal(`Edit ${NAME}`).exists(),
```
cowboyd, line 91: This is redundant, since the following lines are implicitly within the modal context, and if the modal is not there, it will contain all error messaging of the missing modal.

cypress/integration/remote-storage/configurations/crud.spec.js
```diff
@@ -0,0 +1,127 @@
...
+        Modal(`Edit ${NAME}`).find(Button('Cancel')).exists(),
+        Modal(`Edit ${NAME}`).find(Button('Save')).exists(),
```
cowboyd, line 93: This line is redundant since the very next action uses the button.

cypress/integration/remote-storage/configurations/crud.spec.js
```diff
@@ -0,0 +1,127 @@
...
+        Modal(`Remove ${NAME_EDITED}`).find(Button('Cancel')).exists(),
+        Modal(`Remove ${NAME_EDITED}`).find(Button('Delete')).exists(),
+      ])
+
+      .do(Button('Delete').click())
+      .expect([
+        Modal(`Remove ${NAME_EDITED}`).absent(),
+        MultiColumnListCell(NAME_EDITED).absent(),
+      ]);
+  });
```
cowboyd, line 126: optional, but you can probably get away without using the `NAME_EDITED` constant so much by using a matcher. Also, I've removed the redundant interactions.

```suggestion
    const modal = Modal(including('Remove '));
    cy.do(Button('Actions').click())
      .do(Button('Delete').click())
      .expect(modal.exists())
      .do(Button('Delete').click())
      .expect([
        modal.absent(),
        MultiColumnListCell(NAME_EDITED).absent(),
      ]);
  });
```

cowboyd (commented):
🥳 Looking good! 

I've pointed out a bunch of places where there are a lot of redundant statements that don't take into account that interactors always validate the entire reference chain before each action, including all nesting contexts https://frontside.com/blog/2020-07-16-the-lesson-of-bigtest-interactors/

Also, I think that by using matchers you can make the test a bit more readable, and also a little less redundant. There are a couple of places where you could do this, but I only pointed it out on the "Delete" path.

cypress/integration/remote-storage/configurations/crud.spec.js
cowboyd: @axelhunn It should because `Modal(including('Remove'))` is not a cypress command, it is just an expression that returns a data structure. In other words, `Modal()` is a pure function.

It isn't until you say `modal.exists()` or `modal.click()` that anything command-like happens. In fact, one of the suggestions that I was going to make, but forgot to was that you could actually create "save" and "delete" interactors for example at the very top of your file and then use them throughout the tests:

```js
const saveButton = Button('Save');
const deleteButton = Button('Delete');
```

You can now use these everywhere. So, for example, we could re-write the above as:

```js
    const modal = Modal(including('Remove '));
    cy.do([
      Button('Actions').click(),
      deleteButton.click()),
      modal.find(deleteButton.click())
    ])
      .expect([
        modal.absent(),
        MultiColumnListCell(NAME_EDITED).absent(),
      ]);
  });
```

axelhunn: Thanks, it's a great thing!

cypress/integration/remote-storage/configurations/crud.spec.js
```diff
@@ -0,0 +1,127 @@
+import { bigtestGlobals } from '@bigtest/globals';
+import { Link } from '@bigtest/interactor';
```
cowboyd, line 2: Should we re-export `Link` interactor from stripes testing?

axelhunn: I don't see it there...

cowboyd: @JohnC-80 ☝🏻 

Is making `Link` a stripes interactor a good idea so that nobody needs to import directly from `@bigtest/interactor`?

cypress/integration/remote-storage/configurations/crud.spec.js
axelhunn: Will `.find()` API work this way, outside the `do` or `expect` command?

I mean can I do it like that:
```js
const pane = Pane(NAME);
const button = pane.find(Button('Actions'));

cy.do(button.click())
```

cowboyd: It will indeed!

PR #115, opened 2021-07-27
UITEST-81 export mocks for some common modules

zburke:
Export mocks of some common modules, i.e. those provided by a platform to be consumed by UI apps as peers, so apps do not need to maintain their own copies. 

[UITEST-81](https://issues.folio.org/browse/UITEST-81)


zburke:
In the end, this just feels too risky: changes here have the potential to break tests in repositories that are totally disconnected from this one. stripes-testing should provide a stable interactors API, but mocks should remain local.

PR #116, opened 2021-07-29
FAT-901 Refactor Remote Storage Configurations CRUD test using RTL

axelhunn:


cypress/support/index.js
```diff
@@ -1,4 +1,4 @@
-import '@bigtest/cypress';
+import '@testing-library/cypress/add-commands';
```
axelhunn, line 1: I had to remove 
```js
import '@bigtest/cypress';
```
because it broke the `.should(chainers)` assertions like
```js
cy.findByRole('dialog').should('not.exist');
```

axelhunn: ⚠️ This breaks the tests that use interactors!

cowboyd: Interesting. How did it break `.should(chainers)`? In other words, what is the expected behavior, and what is the actual behavior?

taras: I created an issue to track this https://github.com/thefrontside/bigtest/issues/962

axelhunn: `.should(chainers)` gave me an error, until I removed the Bigtest import here.

Took me considerable time to locate the thing BTW )

taras: Sorry for the trouble and thank you for reporting it. We'll look into this ASAP.

taras: @cowboyd it looks like @minkimcello tracked it down https://github.com/thefrontside/bigtest/issues/962#issuecomment-891153757

taras:
@axelhunn what are your thoughts about using `cy.findByText` instead of `cy.findByRole`. Interactors find by test by default so it would be a more accurate comparison. 

For example, in https://github.com/folio-org/stripes-testing/pull/116/files#diff-b5db36e9bc59576a57e2530535fb7117e34427c97e12d528f6287db2e6ce3569R20

`cy.do(Button('+ New').click())` with Interactors would be `cy.findByText('button', { name: /New/ }).click();`

cypress/support/index.js
minkimcello: I've published a preview package with cypress bumped to `7.7.0` so if anyone would like to give it a try, please update your `package.json` and update the version of `@bigtest/cypress` to `mk_cypr770`:
```
{
  "dependencies": {
    "@bigtest/cypress": "mk_cypr770"
  }
}
```

cypress/integration/remote-storage/configurations/crud.spec.js
```diff
@@ -22,106 +7,99 @@ describe('making CRUD', () => {
...
+    });
+
+    cy.findByRole('textbox', { name: /remote storage name/i }).type(NAME);
+    cy.findByRole('combobox', { name: /provider name/i }).select(PROVIDER);
+    cy.findByRole('button', { name: /save/i }).click();
+
+    cy.findAllByRole('dialog').find(':contains(Are you sure)').within(() => {
+      cy.findByRole('button', { name: /cancel/i }).should('be.enabled');
+      cy.findByRole('button', { name: /save/i })
+        .should('be.enabled')
```
cowboyd, line 36: Note that interactors implicitly check to ensure that a button is enabled in order to run actions against it.

cypress/support/index.js
taras: @minkimcello should cypress be a peerDependency for `@bigtest/cypress`?

minkimcello: @taras I think it should be. I only have it as a dependency because someone thought users should only have to install `@bigtest/cypress` instead of `@bigtest/cypress` _and_ `cypress`.

taras: @minkimcello yeah, it looks like it would have prevented this issue

cowboyd:
@taras I think that looking up by role is actually pretty comparable. The difference is that:

1. it is string rather than function based. E.g. `{ role: 'button' }` vs `Button()`
2. rather than being a convention, it is a contract maintained by the component library.
3. interactors implicitly contain custom actions that may be specific to their role. E.g. `accordion.open()`
4. interactors can contain implicit filters that add a semantic layer on top of the HTML. E.g. `Cell({ column: 'price', row: 1 })`

In fact, one way to think about interactors might be as "smart roles"

cypress/support/index.js
cowboyd: @axelhunn @taras @minkimcello 

I've tracked the core issue down to a problem with the way cypress is consuming modules internally. My guess is that it has to do with how cypress was historically bundled and written in coffee script.

https://github.com/cypress-io/cypress/issues/17564 


However, given the number of open issues on the cypress repository (`1.5k`) I don't think we can count on it getting attention any time soon. That means we either need to rename the `expect` method to something like `assert`, or try to do some deep magic workaround.

PR #118, opened 2021-09-16
Add User deletion tests

poflankov:


cypress/integration/users/delete-users.spec.js
```diff
@@ -0,0 +1,72 @@
+import {
+  Button,
+  including,
+  Pane,
+} from '../../../interactors';
+
+describe('Deleting user', () => {
+  const lastName = 'Test123' + Number(new Date()).toString();
+  const ResultsPane = Pane({ index: 2 });
+  const ActionsMenu = ResultsPane.find(Button({ className: including('actionMenuToggle') }));
```
NikitaSedyx, line 10: what about new interactor for ActionsMenu?

cypress/integration/users/delete-users.spec.js
```diff
@@ -0,0 +1,72 @@
...
+        });
+      });
+  });
+
+  it('should be possible by user delete action', function () {
+    const { id } = Cypress.env('user');
+
+    cy.visit(`/users/preview/${id}`);
+    cy.do([
+      ActionsMenu.click(),
```
NikitaSedyx, line 50: it can be common command open menu and press action

cypress/integration/users/delete-users.spec.js
poflankov: I'm afraid I'm not so experienced with interactors yet

cypress/integration/users/delete-users.spec.js
poflankov: I can not imagine how this command could look like. Should it get the action as an argument?

cypress/integration/users/delete-users.spec.js
cowboyd: @poflankov have a look here. https://github.com/folio-org/stripes-testing/blob/master/doc/interactors.md 


Also, for a general introduction to interactors https://frontside.com/interactors/docs

cypress/integration/users/delete-users.spec.js
```diff
@@ -67,6 +68,6 @@ describe('Deleting user', () => {
     ]);
 
     cy.visit(`/users/preview/${id}`);
-    cy.expect(ResultsPane.has({ id: 'pane-userdetails' }));
+    cy.expect(KeyValue().exists({ value: lastName }));
```
cowboyd, line 71: `exists()` does not take any arguments. Instead it checks that an element matching your locator and filters exists.

 In the case of `KeyValue()` without arguments, there are no locator or filters, and `exists()` takes no arguments, so as-is, this will only assert that there is _some_ `KeyValue` present. Any `KeyValue` will satisfy it. 

Instead, I would locate a specific key value and then use `has()` to make an assertion on its expected value: E.g.

```js
    cy.expect(KeyValue('Last Name').has({ value: lastName }));
```

Implicit in this assertion is the assertion that `KeyValue('Last Name').exists()`

cowboyd: (I don't know what the key is for your `KeyValue`, `'Last Name'` was just a guess

poflankov: fixed it

cowboyd: @poflankov you actually found a bug in the docs! Thanks!

cowboyd: Fix is here https://github.com/thefrontside/interactors/pull/84

cypress/integration/users/delete-users.spec.js
poflankov: I tried to make the ActionMenu not a separate interactor, but the Pane interactor method.
Please take a look.

cypress/integration/users/delete-users.spec.js
```diff
@@ -0,0 +1,233 @@
...
+          itemId: Cypress.env('items')[0].id,
+          requesterId: userId,
+          pickupServicePointId: Cypress.env('servicePoints')[0].id,
+          requestDate: '2021-09-20T18:36:56Z',
+        });
+      })
+      .then(() => {
+        verifyUserDeleteImpossible(userId);
+
+        cy.putRequest({
```
NikitaSedyx, line 102: why put when you need to clean up data you created?

cypress/integration/users/delete-users.spec.js
```diff
@@ -0,0 +1,233 @@
...
+          cancelledByUserId: userId,
+          cancellationReasonId: Cypress.env('cancellationReasons')[0].id,
+          cancelledDate: '2021-09-30T16:14:50.444Z',
+        });
+      });
+  });
+
+  it('should be unable in case the user has open loans', function () {
+    const ITEM_BARCODE = Number(new Date()).toString();
+    // const { id: userId, barcode: userBarcode } = Cypress.env('user');
```
NikitaSedyx, line 114: plz remove unnecessary comments

cypress/integration/users/delete-users.spec.js
```diff
@@ -0,0 +1,233 @@
...
+        cy.createItemCheckout({
+          itemBarcode: ITEM_BARCODE,
+          userBarcode: user.barcode,
+          servicePointId: servicePoint.id,
+        });
+      })
+      .then(() => {
+        verifyUserDeleteImpossible(user.id);
+
+        cy.postItemCheckin({
```
NikitaSedyx, line 160: the same

cypress/support/api/blocks.js
```diff
@@ -0,0 +1,19 @@
+Cypress.Commands.add('postBlock', (block) => {
```
NikitaSedyx, line 1: mb names like createBlockHttp, createRequestHttp are better

NikitaSedyx: or createRequestApi - to avoid overrides by commands that use interactors like createCreate (via form)

poflankov: I'm not sure I have a clear vision of the command naming here.
For example we use aforementioned `putRequest` to actually cancel the request.
But if I name the command `cancelRequest`, it could not be reused for another possible Request modifications (which is PUT for)

cypress/support/api/proxy.js
```diff
@@ -0,0 +1,26 @@
...
+});
+
+Cypress.Commands.add('deleteProxyApi', (proxyId) => {
+  cy
+    .okapiRequest({
+      method: 'DELETE',
+      path: `proxiesfor/${proxyId}`,
+    });
+});
+
```
NikitaSedyx, line 20: please remove extra lines

PR #122, opened 2021-10-06
Stcom 886 autosuggest avatar badge interactor

JohnC-80:
Added interactors  and documentation for
- AutoSuggest
- Avatar
- Badge

Ended up borrowing some utility functions from the interactor library for doing just short of what 'fillIn' commonly does (only needed to fill in a filter value for the AutoSuggest, rather than blur the input.

### DIFF
--- a/interactors/autosuggest.js
+++ b/interactors/autosuggest.js
@@ -1,14 +1,36 @@
 import { TextField } from '@interactors/html';
 import HTML from './baseHTML';
+import { dispatchInput, dispatchKeyDown, dispatchKeyUp, setValue, guessCode } from './helpers';
 
 const open = (el) => el.getAttribute('aria-expanded') === 'true';
 
+const fillInput = (element, value) => {
+  const input = element.querySelector('input');
+  input.focus();
+  for (const letter of value) {
+    if (dispatchKeyDown(input, { key: letter, code: guessCode(letter) })) {
+      // don't change the value if the keydown event was stopped
+      setValue(input, input.value + letter);
+      // input is not dispatched if the keydown event was stopped
+      dispatchInput(input, { inputType: 'insertText', data: letter });
+    }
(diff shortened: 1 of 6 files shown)


interactors/autosuggest.js
```diff
@@ -0,0 +1,58 @@
...
+  })
+  .actions({
+    fillIn: ({ find }, value) => find(TextField()).fillIn(value),
+    enterFilter: ({ perform }, value) => perform((el => fillInput(el, value))),
+    select: async (interactor, value) => {
+      if (interactor.is({ open: false })) {
+        await interactor.perform(clearInput);
+        await interactor.perform(el => fillInput(el, value));
+      }
+      await interactor.find(AutoSuggestOption(value)).click();
```
cowboyd, line 56: Nice usage of a filter in an action!

interactors/badge.js
```diff
@@ -0,0 +1,11 @@
+import HTML from './baseHTML';
+
+const color = (element) => element.className;
+
+export default HTML.extend('badge')
+  .selector('[class^=badge-]')
+  .locator(element => element.querySelector('[class^=label-]').textContent)
+  .filters({
+    color,
+    value: (element) => element.querySelector('[class^=label-]').textContent
```
cowboyd, line 10: Are you sure you want `textContent` here and not `innerText`? `textConent` will also include the content of script tags and stylesheets, as opposed to only those things which the browser renders as text.

JohnC-80: Ah.. yes - I like the emphasis on 'the text the browser renders' - it was in a handful of the interactors we had, so I went with it - I might have thought for a half-second to look up the difference. Good call!
Seems like it could be used/preferred across the board...

cowboyd: Yeah, we converted all of ours to use `innerText` and haven't had any issue with it thus far.

PR #123, opened 2021-10-08
STSMACOM-886 repeatable-field-interactor

JohnC-80:


interactors/repeatablefield.js
```diff
@@ -0,0 +1,47 @@
...
+// find by legend
+export const RepeatableFieldInteractor = HTML.extend('repeatable field')
+  .selector('fieldset[data-test-repeatable-field')
+  .locator(el => el.querySelector('legend').textContent)
+  .filters({
+    id: (el) => el.id,
+    emptyMessage: (el) => el.querySelector('[class^=emptyMessage]').textContent,
+    deleteButton: (el) => !!el.querySelector('[data-test-repeatable-field-remove-item-button]'),
+    addButton: (el) => !!el.querySelector('[data-test-repeatable-field-add-item-button]'),
+    addButtonId: (el) => el.querySelector('[data-test-repeatable-field-add-item-button]').id,
```
cowboyd, line 25: Are these necessary to expose? Usually I try to keep the HTML API of an interactor opaque. That means the underlying implementation of the component can change, but the tests don't have too.

interactors/repeatablefield.js
```diff
@@ -0,0 +1,47 @@
...
+      },
+      default: false
+    },
+    removeDisabled: (el) => el.querySelector('[data-test-repeatable-field-remove-item-button]').disabled,
+    itemCount: (el) => el.querySelectorAll('li').length,
+    headLabels: (el) => !!el.querySelector('[data-test-repeatable-field-list-item-labels]'),
+  })
+  .actions({
+    clickAddButton: ({ find }) => find(RepeatableFieldAddButtonInteractor()).click(),
+    clickRemoveButton: ({ perform }, index) => {
```
cowboyd, line 40: Again, I would consider mentioning only the action and not how it's implemented.
```suggestion
    clickAdd: ({ find }) => find(RepeatableFieldAddButtonInteractor()).click(),
    clickRemove: ({ perform }, index) => {
```

PR #127, opened 2021-10-14
STSMACOM-538 refine repeatableField interactor.

JohnC-80:
Removed some implementation-specific naming as well as some more implementation-specific filters, leaving those up to separate interactors at the test level.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "lint": "eslint ."
   },
   "dependencies": {
-    "@interactors/html": "^0.32.0",
+    "@interactors/html": "^0.33.0",
     "date-fns": "^2.16.1",
     "debug": "^4.0.1",
     "element-is-visible": "^1.0.0",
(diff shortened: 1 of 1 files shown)


zburke (approved):
What @cowboyd said: nicely done!

interactors/repeatablefield.js
```diff
@@ -33,11 +30,10 @@ export const RepeatableField = HTML.extend('repeatable field')
...
-    headLabels: (el) => !!el.querySelector('[data-test-repeatable-field-list-item-labels]'),
   })
   .actions({
-    clickAddButton: ({ find }) => find(RepeatableFieldAddButton()).click(),
-    clickRemoveButton: ({ perform }, index) => {
+    clickAdd: ({ find }) => find(RepeatableFieldAddButton()).click(),
+    clickRemove: ({ perform }, index) => {
       return perform((el) => {
         const button = [...el.querySelectorAll('button[data-test-repeatable-field-remove-item-button]')].filter((b, i) => i === index)[0];
         return button ? button.click() : '';
```
cowboyd, line 39: It just occured to me totally after the fact that if you made a `RepeatableFieldRemoveButton` with an `index` filter similar to the way we  did `MultiColumnList` https://github.com/folio-org/stripes-testing/blob/master/interactors/multi-column-list.js#L32

You could have the same API: `field.clickRemove(2)` but you would get much better error messaging and the JS might be a bit cleaner since it would look like:

```js
{
  //...
  clickRemove: ({ find }, index) => find(RepeatableFieldRemoveButton({ index })).click()
}

PR #130, opened 2021-10-20
C675: Create new organization record

poflankov:


cypress/support/index.js
```diff
@@ -11,5 +12,6 @@ import './checkout';
 import './eholdings';
 import './inventory';
 import './users';
+import './organizations';
```
NikitaSedyx, line 15: let's keep alphabetical order

cypress/support/organizations.js
```diff
@@ -0,0 +1,16 @@
+import { Button, Select, TextField } from '../../interactors';
+
+Cypress.Commands.add('createOrganization', ({ name, code, status }) => {
+  const saveButton = Button({ id: 'organization-form-save' });
+
+  cy.expect(saveButton.has({ disabled: true }));
```
NikitaSedyx, line 6: not sure we need expectation that save button is enabled/disabled in createOrg command

zburke:
@poflankov , @NikitaSedyx , what is "C675"? This PR has no description whatsoever and C675 doesn't look like a Jira reference. Help?

zburke:
@poflankov, @NikitaSedyx, Ah, found it, it's a TestRails case linked to https://issues.folio.org/browse/FAT-856. In the future, please name PRs for the Jiras they correspond to since most devs do not have access to TestRails. It's also polite to include a link to the Jira in the description.

@antoneme, it would be great to have more detail in the Jira descriptions since, as above, TestRails is not available to most folks. Additionally, can/should these e2e tickets be moved/cloned from FAT to UITEST, the Jira project corresponding to stripes-testing, so we can track which tests are implemented in each release?

PR #135, opened 2021-11-09
example of page fragments usage

dmitriyfolio:


cypress.json
```diff
@@ -1,8 +1,12 @@
 {
-  "baseUrl": "https://folio-snapshot.dev.folio.org",
+  "baseUrl": "http://localhost:3000/",
```
KyryloBrener, line 2: Do we need these changes here?

dmitriyfolio: no. And may be it's good to add config file example into md file and add such files into .gitignore. Need to discuss with the all Team.

cypress/plugins/dateTools.js
```diff
@@ -0,0 +1,6 @@
+
+export function getCurrentDate(){
```
NikitaSedyx, line 2: it's not plugin, just util

NikitaSedyx:
plz fix linter issues + EOF

cypress.json
dmitriyfolio: I will revert this change at the end of review, before merge in master

cypress/plugins/dateTools.js
dmitriyfolio: we don't have such folder. Do you mean I should add new folder utils in the root?

cypress.json
dmitriyfolio: done

cypress/plugins/dateTools.js
dmitriyfolio: moved to utils in support

KyryloBrener (approved):

MaksimTerekhov (approved):

PR #140, opened 2021-11-17
FAT-734 test for Inventory verifies action menu options after searching and selecting result

MaksimTerekhov:


cypress/integration/inventory/actions-menu.spec.js
```diff
@@ -20,4 +21,18 @@ describe('ui-inventory: actions', () => {
...
+      Actions.saveCQLQueryOption,
+      Actions.exportMARCOption,
+      Actions.showSelectedRecordsOption,
+    ];
+
+    cy.do(InventorySearchPanel.effectiveLocationInput.select([InventorySearchPanel.effectiveLocationValues.mainLibrary]))
+      .then(() => InventorySearchPanel.selectAllCheckboxResult())
+      .then(() => Actions.open())
+      .then(() => optionsShouldEnabled.forEach(option => Actions.optionIsDisabled(option, false)));
+  });
```
JohnC-80, line 37: Why not use interactors directly for the interaction/assertion rather than use 'fragments'?
```
cy.do([MultiSelect('Effective Location').select('Main Library'),
       Button('Select All').click(),
       Button('Actions').click()]);
cy.expect([Action('Save UUID').is({ enabled: true }),
           Action('Save CQL Query').is({ enabled: true }),
           Action('Export Marc').is({ enabled: true }),
           Action('Show Selected Records').is({ enabled: true })]);
```

JohnC-80 (commented):
Just a note to ask if we're putting interactors to best use here...Thanks!

cypress/integration/inventory/actions-menu.spec.js
MaksimTerekhov: I don't use this because names or ids can be changed in the future, so we decided to use fragment for avoid this
I will try to fix this my strange code with many "then" :)

cypress/integration/inventory/actions-menu.spec.js
```diff
@@ -1,23 +1,36 @@
...
-      Actions.showSelectedRecordsOption
-    ];
+    cy.do(Actions.open());
 
-    Actions.open()
-      .then(() => {
-        optionsShouldDisabled.forEach(option => Actions.optionIsDisabled(option, true));
-      });
+    cy.expect([
+      Actions.saveUUIDsOption().is({ disabled: true }),
```
dmitriyfolio, line 15: As I see test related with initial state of Inventories list. Need to add assertion on void list of inventories. May be there are some additional cases, for example, when inventories presented but not selected(please, review this point once again please, I don't know all specifics)

cypress/integration/inventory/actions-menu.spec.js
```diff
@@ -1,23 +1,36 @@
...
+      Actions.saveCQLQueryOption().is({ disabled: true }),
+      Actions.exportMARCOption().is({ disabled: true }),
+      Actions.showSelectedRecordsOption().is({ disabled: true }),
+    ]);
+  });
+
+  it('C196753 verifies action menu options after searching and selecting result', () => {
+    cy.do([
+      InventorySearchPanel.effectiveLocationInput.select([InventorySearchPanel.effectiveLocationValues.mainLibrary]),
```
dmitriyfolio, line 24: lets wrap this to  InventorySearchPanel.searchByEffectiveLocation(specialLocation)? And you can specify mainLibrary as default parameter value in this test. We will get InventorySearchPanel.searchByEffectiveLocation();
May be we can skip a few words in naming and get just InventorySearch.byEffectiveLocation();

cypress/integration/inventory/actions-menu.spec.js
```diff
@@ -1,23 +1,36 @@
...
+      InventorySearchPanel.firstResultCheckbox().click(),
```
dmitriyfolio, line 25: need to think, but can be useful to implement something like this: 
1. Read lines count
2. Randomly select 1 or 2 items(sometime only 1 row, in another runs - 2 rows) into the list of inventories
3. Check state of options into Actions menu

MaksimTerekhov: For this test it is excess. We need to check only action options, not list of inventory.

cypress/integration/inventory/actions-menu.spec.js
MaksimTerekhov: If inventories presented but not selected It means something went wrong and test failed. It is ok.

cypress/support/fragments/inventory/inventorySearchPanel.js
```diff
@@ -0,0 +1,13 @@
+import { Checkbox, MultiColumnListCell, MultiSelect } from '../../../../interactors';
+
+export default class InventorySearchPanel {
+  static effectiveLocationInput = MultiSelect({ 'id': 'multiselect-6' });
+  static effectiveLocationValues = {
+    'mainLibrary': 'Main Library'
+  }
+  static #firstResultCheckbox = MultiColumnListCell({ 'row': 0, 'columnIndex': 0 }).find(Checkbox());
```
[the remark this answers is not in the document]
MaksimTerekhov: what do you mean? If you talk about finding random checkbox, It is excess for this test. We need any checkbox for action options checks

cypress/integration/inventory/actions-menu.spec.js
MaksimTerekhov: Yes, naming so bad. I will think how it can be fixed

cypress/support/fragments/inventory/inventorySearchPanel.js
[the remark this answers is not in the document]
dmitriyfolio: image
this one

cypress/integration/inventory/actions-menu.spec.js
dmitriyfolio: as I see in test precondition: "and the results pane is empty.". I can't see evident validation of this requirement in your test. Such additional check can get you answer about the reason of failed test quicker. It is one more benefit of my suggestion.

cypress/support/fragments/inventory/actions.js
```diff
@@ -2,19 +2,29 @@ import { Button } from '../../../../interactors';
 
 export default class Actions {
```
dmitriyfolio, line 3: could you rename Actions to InventoryActions?

cypress/support/fragments/inventory/inventorySearch.js
```diff
@@ -0,0 +1,22 @@
...
+    'mainLibrary': 'Main Library'
+  }
+
+  static #firstResultCheckbox = MultiColumnListCell({ 'row': 0, 'columnIndex': 0 }).find(Checkbox());
+
+  static getFirstResultCheckbox() {
+    return this.#firstResultCheckbox;
+  }
+
+  static getResults() {
```
dmitriyfolio, line 15: could you add jsdoc to this method?

cypress/integration/inventory/actions-menu.spec.js
```diff
@@ -1,23 +1,38 @@
...
+      InventoryActions.saveUUIDsOption(),
+      InventoryActions.saveCQLQueryOption(),
+      InventoryActions.exportMARCOption(),
+      InventoryActions.showSelectedRecordsOption()
+    ]);
+  });
+
+  it('C196753 verifies action menu options after searching and selecting result', () => {
+    cy.do([
+      InventorySearch.byEffectiveLocation([InventorySearch.effectiveLocation.mainLibrary]),
```
dmitriyfolio, line 26: I suggest to use InventorySearch.effectiveLocation.mainLibrary as default value. In this case you will have just InventorySearch.byEffectiveLocation(). And you can change this to extra wrapper with name  InventorySearch.getNotVoidResult() , which will have request of InventorySearch.byEffectiveLocation

dmitriyfolio (approved):

KyryloBrener (approved):

cypress/support/fragments/inventory/inventorySearch.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  static getSearchResults() {
+    return MultiColumnList();
+  }
+
+  static byEffectiveLocation(values = [this.effectiveLocation.mainLibrary]) {
+    return InventorySearch.#effectiveLocationInput.select(values);
+  }
+}
```
cowboyd, line 23: Since all of these are just static constants, why use a class here at all? Why not just

```js
export default {
    effectiveLocationInput: MultiSelect({ 'id': 'multiselect-6' }),
    effectiveLocation:  {
      'mainLibrary': 'Main Library'
    },
    firstResultCheckbox: MultiColumnListCell({ 'row': 0, 'columnIndex': 0 }).find(Checkbox()),
 
    getFirstResultCheckbox: get() {
      return this.firstResultCheckbox;
    },
 
    getSearchResults() {
      return MultiColumnList();
    }
 
    byEffectiveLocation(values = [this.effectiveLocation.mainLibrary]) {
      return InventorySearch.#effectiveLocationInput.select(values);
    }
}
```

But honestly, that begs the question, why not just make these individual exports? E.g.

```js
export const effectiveLocationInput = MultiSelect({ 'id': 'multiselect-6' });
```

cowboyd: Then, you can import a single symbol if that's all you're going to use.

```js
import { effectiveLocationInput } from '.../fragments/inventory/inventorySearch`;
```

or if you want to slurp the whole module, and use it as before:

```js
import * as InventorySearch from '../fragments/inventory/inventorySearch';

InventorySearch.effectiveLocationInput
```

dmitriyfolio: in case of class usage code will have more special name like InventorySearch.byEffectiveLocation() instead of just byEffectiveLocation(). Class wraps all related with special page object into one place

cowboyd: > code will have more special name like InventorySearch.byEffectiveLocation() instead of just byEffectiveLocation().

As my code example shows though, you don't need classes to do this. The class syntax is just noise. Also, if you export standalone functions, you can still use the `InventorySearch.byEffectiveLocation()`. It's more flexible because you have the option to do both.

PR #141, opened 2021-11-22
stabilization + the rest smoke tests with Notes

dmitriyfolio:
update of packages
fix issue with timeouts of interactors
create common long delay object in cypress
one more test about Notes management


cypress.json
```diff
@@ -3,12 +3,13 @@
...
-  "defaultCommandTimeout": 11000,
+  "defaultCommandTimeout": 21000,
   "pageLoadTimeout": 120000,
   "env": {
     "OKAPI_HOST": "folio-snapshot.dev.folio.org",
     "OKAPI_TENANT": "diku",
     "diku_login": "diku_admin",
-    "diku_password": "admin"
+    "diku_password": "admin",
+    "long_delay": 20000
```
MaksimTerekhov, line 13: What is it? Where it can be usable? I can't find it in official docs:
https://docs.cypress.io/guides/references/configuration#Node-version

cypress/support/utils/cypressTools.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable import/prefer-default-export */
+export function getLongDelay() {
```
MaksimTerekhov, line 2: May be better to create class instead of funtion and describe all our timeouts like default, long or may be others

cypress/support/login.js
```diff
@@ -11,13 +12,18 @@ Cypress.Commands.add('login', (username, password) => {
...
-  cy.do([
-    TextField('Username').fillIn(username),
-    TextField('Password').fillIn(password),
-    Button('Log in').click(),
-  ]);
+  // Todo: find the way to wrap interactor to cy chainable object
+  cy.get('input[name=username]', getLongDelay()).should('be.visible');
+  cy.do(TextField('Username').fillIn(username));
+  cy.do(TextField('Password').fillIn(password));
+  cy.do(Button('Log in').click());
```
MaksimTerekhov, line 19: Why it was replaced? Previous verison had chainable cy object and it worked

cypress/support/login.js
```diff
@@ -11,13 +12,18 @@ Cypress.Commands.add('login', (username, password) => {
...
+
+  // TODO: find the way how customize waiter timeout in case of interactors(cy.wrap may be)
```
MaksimTerekhov, line 21: _defaultCommandTimeout_ in _cypress.json_ help us with it. If it need we can set new value of this parametr. Why we need new ways for expecting elements?

package.json
```diff
@@ -28,10 +28,10 @@
   "devDependencies": {
     "@babel/eslint-parser": "^7.16.3",
     "@folio/eslint-config-stripes": "^5.0.0",
-    "@interactors/with-cypress": "^0.1.3",
+    "@interactors/with-cypress": "^0.1.4",
     "babel-eslint": "^10.1.0",
     "bigtest": "^0.14.0",
-    "cypress": "^9.0.0",
```
MaksimTerekhov, line 34: Everythin works with new versions? Version increments can be not easy and a lot of fixed may be need

cypress.json
dmitriyfolio: you are right. It's piece of old code, will be removed from this PR

cypress/support/utils/cypressTools.js
dmitriyfolio: I can't see benefits from class wrapping. Could you describe the reason?
" describe all our timeouts like default, long or may be others" - can't understand you, give an example please

cypress/support/login.js
dmitriyfolio: if we would say about regular situations - yes, we can work through defaultCommandTimeout. From other side, some set actions can be complicated and related with long responses from application(for example, it's required in some test to create object with maximum volume of test data). For such cases I suggest to specify additional delay through getLongDelay(), which will additionally indicate to us such special places in tests and fragments.

package.json
dmitriyfolio: my tests work. Could you provide your stable tests?

cypress/support/utils/cypressTools.js
MaksimTerekhov: Now we save our timeouts in different places,
For example one timeout in cypress.json another timeout in cypressTools
I suggest to use one place for timeouts
May be this is not a class, may be use another way for it

cypress/support/login.js
MaksimTerekhov: ok, I understood, thanks

package.json
MaksimTerekhov: Now I don't. I mean if you increment some dependencies, we should get some benefits from new version and we should sure about other functional will work. If you don't sure of it. may be we don't need increment now

cypress/support/utils/cypressTools.js
dmitriyfolio: additionally we have timeout related with interactors. For me - it's ok for now and may be turned in the future(for example as set of env variables)

dmitriyfolio: I've found that we use in our code different places with different long timeouts(you can see it into original code in this PR). And my change related with wrapping long timeout into one place(function) and continue to use it without doubts which value should be specified if it's needed.

MaksimTerekhov: ok, I agree with you to think about it in future

package.json
dmitriyfolio: I see a lot of cypress strange things and hope each new version will get us more convenient and effective tool. My suggestion - if our tests still stable - we can up our dependencies as we wish.  And of cause we have the ability fix version in case of known issues/blockers.

cypress.json
dmitriyfolio: removed

cypress/support/login.js
dmitriyfolio: reverted

cypress/support/utils/cypressTools.js
dmitriyfolio: thank's

package.json
MaksimTerekhov: In this case I don't see problems because this you increment minor version:
https://docs.cypress.io/guides/references/changelog#9-1-0
But in future I suggest to create separate branch for it and check that all tests works

dmitriyfolio: correct, you can see such requirement into draft of code review process -" If you change existing packages versions, run all existing tests to verify that already implemented tests still work" - https://wiki.folio.org/display/DQA/Code+review

MaksimTerekhov (approved):

KyryloBrener (approved):

PR #143, opened 2021-11-29
FAT-735 test for verifing export UUIDs instances

MaksimTerekhov:


cypress/integration/inventory/export.spec.js
```diff
@@ -0,0 +1,40 @@
...
+describe('inventory / actions: export UUIDs', () => {
+  beforeEach('navigates to Inventory and delete downloads folder', () => {
+    cy.login(Cypress.env('diku_login'), Cypress.env('diku_password'));
+    cy.visit(TopMenu.inventoryPath);
+  });
+
+  it('C9284 verifies export UUIDs instances', () => {
+    cy.do([
+      InventorySearch.searchByEffectiveLocation(),
+      InventoryActions.open(),
```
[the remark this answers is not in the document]
dmitriyfolio: and 3000 can be moved into env variables

cypress/integration/inventory/export.spec.js
```diff
@@ -0,0 +1,40 @@
+import TopMenu from '../../support/fragments/topMenu';
+import InventorySearch from '../../support/fragments/inventory/inventorySearch';
+import InventoryActions from '../../support/fragments/inventory/inventoryActions';
+import FileManager from '../../support/utils/fileManager';
+
+
+describe('inventory / actions: export UUIDs', () => {
+  beforeEach('navigates to Inventory and delete downloads folder', () => {
```
dmitriyfolio, line 8: We don't need to clear test folder before/after test now, am I right?

cypress/support/fragments/inventory/inventoryActions.js
```diff
@@ -19,4 +19,15 @@ export default {
       cy.expect(element.is({ disabled: false }));
     });
   },
+  verifySaveUUIDsFileName(actualName) {
```
dmitriyfolio, line 22: "The time stamp is local time of the user running the test" - can't see this check

cypress/support/fragments/inventory/inventoryActions.js
```diff
@@ -19,4 +19,15 @@ export default {
...
+    const expectedFileNameMask = /SearchInstanceUUIDs\d{4}-\d{2}-\d{2}T\d{2}_\d{2}_\d{2}\+\d{2}_\d{2}\.csv/gm;
+    const isCorrectFilename = expectedFileNameMask.test(actualName);
+    expect(isCorrectFilename).to.eq(true);
```
dmitriyfolio, line 25: What "expect" is? How it works?

cypress/integration/inventory/export.spec.js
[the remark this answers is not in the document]
MaksimTerekhov: > additionally, can we put this code into Inventories.saveInstancesUUIDs():
" cy.wait('@getids').then((obj) => {
const expectedUUIDs = [];
obj.response.body.ids.forEach((elem) => { expectedUUIDs.push(elem.id); });
// Need time for download file TODO: think about how it can be fixed
  cy.wait(3000);"
expectedUUIDs can be return value, which will be processed in next steps of this test

I tried but I couldn't put working with interceptor into separate function. It borns a lof of exception and we need to think about asyn and sync code with this solution. But if I use interseptor in test it so easy works without exceptions

dmitriyfolio (approved):

cypress/integration/inventory/export.spec.js
```diff
@@ -0,0 +1,37 @@
...
+
+describe('inventory / actions: export UUIDs', () => {
+  beforeEach('navigates to Inventory', () => {
+    cy.login(Cypress.env('diku_login'), Cypress.env('diku_password'));
+    cy.visit(TopMenu.inventoryPath);
+  });
+
+  it('C9284 verifies export UUIDs instances', () => {
+    cy.do(InventorySearch.searchByEffectiveLocation());
+    InventoryActions.saveUUIDs();
```
cowboyd, line 15: This is a bare statement, and should be part of the `cy.do`, no?

cypress/integration/inventory/export.spec.js
```diff
@@ -0,0 +1,37 @@
...
+
+    // TODO: think about move it to separate func
+    cy.intercept('/search/instances/ids**').as('getIds');
+    cy.wait('@getIds').then((obj) => {
+      const expectedUUIDs = InventorySearch.getUUIDsFromRequest(obj);
+
+      // Need time for download file TODO: think about how it can be fixed
+      cy.wait(Cypress.env('downloadTimeout'));
```
cowboyd, line 23: As indicated, It's a code-smell to have waits inside our tests. Instead, we should ask ourselves: "how will the user interface reflect the fact that this operation". For example:

```js
// SearchResults
cy.expect(SearchResults().exist())
```

cypress/integration/inventory/export.spec.js
```diff
@@ -0,0 +1,37 @@
...
+        .then((downloadedFilenames) => {
+          const lastDownloadedFilename = downloadedFilenames.sort()[downloadedFilenames.length - 1];
+          InventoryActions.verifySaveUUIDsFileName(lastDownloadedFilename);
+
+          FileManager.readFile(lastDownloadedFilename)
+            .then((actualUUIDs) => {
+              InventoryActions.verifySavedUUIDs(actualUUIDs, expectedUUIDs);
+            });
+        });
+    });
```
cowboyd, line 35: Instead of introducing yet another paradigm, consider just creating new interactors out of existing ones instead of "fragments" They are designed to compose well:

https://frontside.com/interactors/docs/composition

cypress/integration/inventory/export.spec.js
MaksimTerekhov: you right, fixed

cypress/integration/inventory/export.spec.js
MaksimTerekhov: We use interactors inside fragments. We use fragments because we want that tests look shorter and easy for undersandable

cypress/integration/inventory/export.spec.js
MaksimTerekhov: > "how will the user interface reflect the fact that this operation".

UI didn't affect by download process. We just click to action and then download starts

> cy.expect(SearchResults().exist())

I don't see how it can help. When download process starts chrome creates an empty file with id instead of full name and empty inside. This cy.expect can pick this file but I need full file after download ends. And I just don't know ho I can wait while file will downloaded fully. May be if you know, can you help with it please? It may be links or some examples
If you do, I will replace this code

PR #152, opened 2021-12-07
Fat-737 test for verifying selected result for quick export

MaksimTerekhov:


KyryloBrener (approved):

cypress/integration/inventory/export.spec.js
```diff
@@ -37,4 +38,14 @@ describe('inventory / actions: export UUIDs', () => {
         });
     });
   });
+
+  it('C196755 verifies search result counts and selected counts', () => {
```
dmitriyfolio, line 42: Scroll down so that records selected in the first step are not visible anymore  - may be this portion of manual test is required, please check

dmitriyfolio (approved):

cypress/integration/inventory/export.spec.js
MaksimTerekhov: Thanks
But I found it excess for this test because I check alredy ckeck whole string with search and selecter results

PR #155, opened 2021-12-08
Fat 1114 quick marc bib rest tests

dmitriyfolio:


KyryloBrener (approved):
No comments except syntax issues already highlighted automatically.

cypress/support/fragments/inventory/inventoryViewSource.js
```diff
@@ -11,5 +13,6 @@ export default {
   },
   waitLoading: () => {
     cy.expect(HTML(including(title)).exists());
-  }
+  },
+  close: () => cy.do(closeButton.click())
```
MaksimTerekhov, line 17: JFYI, you can use _ instead of () in case without function arguments
`_ => cy.do(closeButton.click())`

of course it is not important, but may be interesting

MaksimTerekhov (approved):

PR #159, opened 2021-12-13
Fat 1115 init quickmarc holding test

dmitriyfolio:


cypress/support/fragments/inventory/inventoryViewSource.js
```diff
@@ -11,8 +12,11 @@ export default {
   notContains:(notExpectedText) => {
     cy.expect(HTML(including(notExpectedText)).absent());
   },
-  waitLoading: () => {
-    cy.expect(HTML(including(title)).exists());
+  waitInstanceLoading: () => {
+    cy.expect(HTML(including(instanceTitle)).exists());
+  },
+  waitHoldingLoading: () => {
+    cy.expect(HTML(including(holdingTitle)).exists());
```
MaksimTerekhov, line 19: May be better to create one function with parameter title?

```
waitLoading: title => {
    cy.expect(HTML(including(title)).exists());
  },
```
or parameter could be page name:

```
waitLoading: pageName => {
    const title = pageName.title
    cy.expect(HTML(including(title)).exists());
  },
```

cypress/support/fragments/inventory/holdingsRecordView.js
```diff
@@ -0,0 +1,34 @@
...
+
+const root = Layer('View holdings record');
+const actionsButton = root.find(Button('Actions'));
+const closeButton = root.find(Button({ icon: 'times' }));
+const editInQuickMarcButton = Button({ id: 'clickable-edit-marc-holdings' });
+const viewSourceButton = Button({ id: 'clickable-view-source' });
+
+export default {
+  newHolding : {
+    rowsCountInQuickMarcEditor : 6
```
MaksimTerekhov, line 14: Why it hardcoded? Is it possible to put this value from test? or is it just a constant?

cypress/support/fragments/inventory/inventoryViewSource.js
dmitriyfolio: in this case you will need to provide right strings in test   or specify somewhere special related constants with reusing into tests.... I can't see benefits from your suggestion, may be I've lost something

cypress/support/fragments/inventory/holdingsRecordView.js
dmitriyfolio: Because I don't have option to read actual rows count into quickmarc editor. I set related TODO into code and will back to this in the future

MaksimTerekhov (approved):

KyryloBrener (approved):

PR #164, opened 2021-12-16
Fat 811

KyryloBrener:


cypress/support/fragments/finance/fiscalYears/fiscalYears.js
```diff
@@ -0,0 +1,70 @@
...
+      Button('Save & Close').click(),
+      TextField('Code*').fillIn('some code'),
+      Button('Save & Close').click(),
+      TextField('Period Begin Date*').fillIn('05/05/2021'),
+      Button('Save & Close').click(),
+      // try to navigate without saving
+      Button('Agreements').click(),
+      Button('Keep editing').click,
+      Button('Cancel').click(),
+      Button('Close without saving').click()
```
MaksimTerekhov, line 45: May be move button to constants? I mean make a variables above export like this:

```
const saveAndClose = Button('Save & Close')
const agreements = Button('Agreements')
```

And in code you can use

```
saveAndClose.click()
agreements.click()
```

In this case we avoid problems with buttons with a lot of same namings and if one of the button will be renamed you need fix name in one line.
What do you think?

cypress/support/fragments/finance/fiscalYears/fiscalYears.js
```diff
@@ -0,0 +1,70 @@
...
+    ]);
+  },
+
+  searchByName : (fiscalYearName) => {
+    cy.do([
+      SearchField({ id: 'input-record-search' }).selectIndex('Name'),
+      SearchField({ id: 'input-record-search' }).fillIn(fiscalYearName),
```
MaksimTerekhov, line 52: What if ID will be changed?
Let's move this interactor to constant above `export default {`
I met same case some days ago. On inventory module some selectors was changed and I am not sure about future others selectors :)

MaksimTerekhov (approved):

TetianaParanich (approved):

PR #191, opened 2022-01-18
UITEST-89 Added AdvancedSearch interactor

BogdanDenis:
## Description
Adding interactor for new `<AdvancedSearch>` component

## Issues
[UITEST-89](https://issues.folio.org/browse/UITEST-89)

## Related PRs
https://github.com/folio-org/stripes-components/pull/1706
https://github.com/folio-org/ui-marc-authorities/pull/57

### DIFF
--- a/interactors/advanced-search.js
+++ b/interactors/advanced-search.js
@@ -1,0 +1,33 @@
+import Button from './button';
+import Select from './select';
+import TextField from './text-field';
+import HTML from './baseHTML';
+
+export const AdvancedSearchRow = HTML.extend('advanced search row')
+  .selector('[class*=AdvancedSearchRow-]')
+  .filters({
+    index: el => parseInt(el.getAttribute('data-row-index'), 10),
+    query: el => el.querySelector('input').getAttribute('value'),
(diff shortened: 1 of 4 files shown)


interactors/advanced-search.js
```diff
@@ -0,0 +1,33 @@
+import Button from './button';
+import Select from './select';
+import TextField from './text-field';
+import HTML from './baseHTML';
+
+export const AdvancedSearchRow = HTML.extend('advanced search row')
+  .selector('[class*=AdvancedSearchRow-]')
```
dmitriyfolio, line 7: can it be more special, something which begins from ^ . Or added hardcoded attribute(id, name, index for example)

dmitriyfolio (approved):

interactors/advanced-search.js
BogdanDenis: I'm not sure what else we can use here... there are 6 rows on the page so we can't use ids.
But you can search a row by index! For example `AdvancedSearchInteractor({ index: 0 })`

mariiaschv (approved):

Svitlana-Korol (approved):

PR #197, opened 2022-01-27
FAT-824 Move multiple items from one holdings to another holdings within an instance

MaksimTerekhov:


cypress/support/fragments/inventory/inventoryInstance.js
```diff
@@ -101,6 +120,43 @@ export default {
...
+  },
+
+  openHoldings: () => {
+    cy.do([
+      firstHolding.clickHeader(),
+      secondHolding.clickHeader()
+    ]);
+  },
+
+  verifySuccessCalloutMovingMessage(count) {
```
KyryloBrener, line 132: I've already created method for verifying callout messages (see fragments/finance/FinanceHelper/checkCalloutMessage(text, calloutType)). I think we can move it to some general helper class and use one verification method in all tests. Maybe not in this PR, in scope of refactoring for example.

KyryloBrener (approved):

cypress/integration/inventory/moving-items.spec.js
```diff
@@ -0,0 +1,23 @@
+import TopMenu from '../../support/fragments/topMenu';
+import InventorySearch from '../../support/fragments/inventory/inventorySearch';
+import testTypes from '../../support/dictionary/testTypes';
+import InventoryInstance from '../../support/fragments/inventory/inventoryInstance';
+
+describe('inventory: moving items', () => {
+  beforeEach('navigates to Inventory', () => {
+    cy.login(Cypress.env('diku_login'), Cypress.env('diku_password'));
```
dmitriyfolio, line 8: test has requirement about user permission. Need to prepare such user with minimal permissions around permission "Move items"

cypress/integration/inventory/moving-items.spec.js
```diff
@@ -0,0 +1,23 @@
...
+    cy.visit(TopMenu.inventoryPath);
+  });
+
+  it('C15185 verifies moving multiple items between holding', { tags: [testTypes.smoke] }, () => {
+    InventorySearch.byEffectiveLocation();
+    cy.do(InventorySearch.getSearchResult().click());
```
dmitriyfolio, line 14: wrapper like selectRecord() can be applicable instead of cy.do(click())

cypress/integration/inventory/moving-items.spec.js
```diff
@@ -0,0 +1,23 @@
...
+import testTypes from '../../support/dictionary/testTypes';
+import InventoryInstance from '../../support/fragments/inventory/inventoryInstance';
+
+describe('inventory: moving items', () => {
+  beforeEach('navigates to Inventory', () => {
+    cy.login(Cypress.env('diku_login'), Cypress.env('diku_password'));
+    cy.visit(TopMenu.inventoryPath);
+  });
+
+  it('C15185 verifies moving multiple items between holding', { tags: [testTypes.smoke] }, () => {
```
dmitriyfolio, line 12: do we have a chance to prepare special test data to this test?

cypress/support/fragments/inventory/inventoryInstance.js
```diff
@@ -16,6 +31,10 @@ const deriveNewMarcBibRecord = Button({ id:'duplicate-instance-marc' });
 const addMarcHoldingRecordButton = Button({ id:'create-holdings-marc' });
 const viewHoldingsButton = Button('View holdings');
 const notesSection = Section({ id: 'instance-details-notes' });
+const moveItemsButton = Button({ id: 'move-instance-items' });
+const firstHolding = Accordion({ label: including('Holdings: Main Library') });
```
dmitriyfolio, line 35: need to organize expected hardcoded values. For example: move it into constants or objects or method parameters with meaning names

cypress/support/fragments/inventory/inventoryInstance.js
```diff
@@ -101,6 +120,43 @@ export default {
...
+  verifySuccessCalloutMovingMessage(count) {
+    cy.expect(callout().is({ textContent: `${count} item has been successfully moved.` }));
+  },
+
+  moveItemToAnotherHolding() {
+    this.openHoldings();
+
+    cy.do([
+      firstHolding.find(MultiColumnListRow()).find(Checkbox()).click(),
+      firstHolding.find(Dropdown({ label: 'Move to' })).choose(including('Annex K1')),
```
dmitriyfolio, line 141: interactor can be moved in constant

cypress/integration/inventory/moving-items.spec.js
dmitriyfolio, line 12: 1. Test in testrail related with Holdings part of inventory. Your test related with Instances. Please double check this
2. I can't see verification of movement(just appearing of message from Folio). It is enough?

MaksimTerekhov: 1. You won't able moving item between holdings inside Holding page, so in test case 'View the holdings' means expand this tabs
image

2. I moved item to another holding, then I return it to previous holding. And if item isn't moved I will not return it to previous holding. I think it is enough for this case with checking success message

cypress/support/fragments/inventory/inventoryInstance.js
MaksimTerekhov: Interactor already in constant:

`const firstHolding = Accordion({ label: including('Holdings: Main Library') });`

In this case I use my constant for checking elements inside

cypress/integration/inventory/moving-items.spec.js
MaksimTerekhov: As I can see, we need to add new holding. It seems difficult and test duration will very increase.

cypress/support/fragments/inventory/inventoryInstance.js
MaksimTerekhov: I agree about future, but now I need a function with hardcoded string about moving items between holdings

MaksimTerekhov: I changed my code and I decided to reuse your function, thx)

dmitriyfolio (approved):

PR #202, opened 2022-02-02
FAT-741 Export small number of instance records - default instance mapping profile

MaksimTerekhov:
https://foliotest.testrail.io/index.php?/cases/view/9288

### DIFF
--- a/cypress/support/fragments/data-export/dataExportResults.js
+++ b/cypress/support/fragments/data-export/dataExportResults.js
@@ -42,8 +42,6 @@ export default {
       resultRow.fileName.perform(element => {
         expect(element.innerText).to.equal(resultFileName);
         expect(element.innerText).to.include(`-${jobId}.mrc`);
-        console.log(element);
-        cy.pause();
       })
     );
 
--- a/cypress/support/fragments/data-export/dataExportResults.js
+++ b/cypress/support/fragments/data-export/dataExportResults.js
@@ -13,5 +13,24 @@ export default {
       }),
       MultiColumnListCell({ 'row': 0, 'column': defaultJobProfile }).exists(),
     ]);
-  }
+  },
+
+  verifySuccessExportResultCells() {
+    const dateString = /\d{1,2}\/\d{1,2}\/\d{4},\s\d{2}:\d{2}\s\w{2}/gm;
+    return cy.do([
+      MultiColumnListCell({ 'row': 0, columnIndex: 1 }).is({ content: 'Completed' }),
+      MultiColumnListCell({ 'row': 0, columnIndex: 2 }).perform(element => {
+        expect(element.innerText).to.match(/\d{1,4}/gm);
+      }),
+      MultiColumnListCell({ 'row': 0, columnIndex: 3 }).is({ content: '' }),
+      MultiColumnListCell({ 'row': 0, columnIndex: 4 }).is({ content: defaultJobProfile }),
+      MultiColumnListCell({ 'row': 0, columnIndex: 5 }).perform(element => {
+        expect(element.innerText).to.match(dateString);
+      }),
+      MultiColumnListCell({ 'row': 0, columnIndex: 6 }).is({ content: 'DIKU ADMINISTRATOR' }),
+      MultiColumnListCell({ 'row': 0, columnIndex: 7 }).perform(element => {
+        expect(element.innerText).to.match(/\d{1,4}/gm);
+      }),
+    ]);
+  },
 };
--- a/cypress/support/fragments/data-export/dataExportResults.js
+++ b/cypress/support/fragments/data-export/dataExportResults.js
@@ -1,36 +1,60 @@
 import InventorySearch from '../inventory/inventorySearch';
 import { MultiColumnListCell } from '../../../../interactors';
+import DateTools from '../../utils/dateTools';
 
-const defaultJobProfile = 'Default instances export job profile';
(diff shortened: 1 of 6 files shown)


KyryloBrener (approved):

cypress/support/fragments/data-export/dataExportResults.js
```diff
@@ -13,5 +13,24 @@ export default {
       }),
       MultiColumnListCell({ 'row': 0, 'column': defaultJobProfile }).exists(),
     ]);
-  }
+  },
+
+  verifySuccessExportResultCells() {
+    const dateString = /\d{1,2}\/\d{1,2}\/\d{4},\s\d{2}:\d{2}\s\w{2}/gm;
+    return cy.do([
+      MultiColumnListCell({ 'row': 0, columnIndex: 1 }).is({ content: 'Completed' }),
```
dmitriyfolio, line 21: can be useful to define constants with interactors and meaning names. It would be simpler to analyze.

cypress/support/fragments/data-export/dataExportResults.js
```diff
@@ -13,5 +13,24 @@ export default {
       }),
       MultiColumnListCell({ 'row': 0, 'column': defaultJobProfile }).exists(),
     ]);
-  }
+  },
+
+  verifySuccessExportResultCells() {
```
dmitriyfolio, line 18: can't see verification of 
"File name. File name should be in bold and **should be clickable**. **The name of the file should contain the name of the input file (with UUIDs) followed by job id**."

cypress/support/fragments/data-export/dataExportResults.js
```diff
@@ -13,5 +13,24 @@ export default {
...
+    const dateString = /\d{1,2}\/\d{1,2}\/\d{4},\s\d{2}:\d{2}\s\w{2}/gm;
+    return cy.do([
+      MultiColumnListCell({ 'row': 0, columnIndex: 1 }).is({ content: 'Completed' }),
+      MultiColumnListCell({ 'row': 0, columnIndex: 2 }).perform(element => {
+        expect(element.innerText).to.match(/\d{1,4}/gm);
```
dmitriyfolio, line 23: "Total:  contains the number of UUIDs in the file that triggered the export" - add implementation of this expectation

cypress/support/fragments/data-export/dataExportResults.js
```diff
@@ -13,5 +13,24 @@ export default {
...
+      }),
+      MultiColumnListCell({ 'row': 0, columnIndex: 3 }).is({ content: '' }),
+      MultiColumnListCell({ 'row': 0, columnIndex: 4 }).is({ content: defaultJobProfile }),
+      MultiColumnListCell({ 'row': 0, columnIndex: 5 }).perform(element => {
```
dmitriyfolio, line 27: as I remember you have experience in implementation of steps like "Ended running: date and time of the completion of the export job"

cypress/support/fragments/data-export/dataExportResults.js
```diff
@@ -13,5 +13,24 @@ export default {
...
+        expect(element.innerText).to.match(dateString);
+      }),
+      MultiColumnListCell({ 'row': 0, columnIndex: 6 }).is({ content: 'DIKU ADMINISTRATOR' }),
+      MultiColumnListCell({ 'row': 0, columnIndex: 7 }).perform(element => {
```
dmitriyfolio, line 31: I hop you can catch job id to finish this step fully. "ID of the export job". And what is the meaning of "The records are correctly formatted"? is it related with format of jobId or all row?

cypress/support/fragments/data-export/dataExportResults.js
MaksimTerekhov: I find string in result table by this condition. You can see it inside function:
`exportWithDefaultInstancesJobProfile`

cypress/support/fragments/data-export/dataExportResults.js
MaksimTerekhov: Do you think it is need this? I think match to regular expression will be ok

dmitriyfolio: such extra check can help to find issue with time resolving

cypress/support/fragments/data-export/dataExportResults.js
dmitriyfolio: I can't see implementation for points:
-  should be clickable
- followed by job id
may be  in bold should be removed from TC in TestRail. Need to check with PO and Team

cypress/support/fragments/data-export/dataExportResults.js
MaksimTerekhov: >  And what is the meaning of "The records are correctly formatted"?

This about whole table formating. And in this tests definetly we don't want to check whole table records

cypress/support/fragments/data-export/dataExportResults.js
MaksimTerekhov: About bold:
It is not possible to check this because I can't get css attribute "font-weight". This property inherited a lot of times and this element don't contain this property inside `document.querySelector().style.fontWeight`

About clickable:
We have a lot of tests in another teams with clicking on this filename. After clicking download .mrc file will start. It seems we have excess checks

dmitriyfolio: you can just check href availability

dmitriyfolio (approved):

PR #216, opened 2022-02-09
Fat 1466 tc 345404

dmitriyfolio:


MaksimTerekhov (approved):

KyryloBrener (approved):

cypress/support/fragments/quickMarcEditor.js
```diff
@@ -207,34 +214,37 @@ export default {
       cy.get(`input[name="${name}"]`).click();
       cy.get(`input[name="${name}"]`).type(`{backspace}{backspace}${tag008HoldingsBytesProperties.specRet0.newValue}`);
     });
-    pressSaveAndClose();
-    HoldingsRecordView.waitLoading();
+    this.pressSaveAndClose();
```
BogdanDenis, line 217: This won't work correctly because you can't have `this` in a `static` method

cypress/support/fragments/quickMarcEditor.js
```diff
@@ -207,34 +214,37 @@ export default {
...
-  clearTag008:() => {
+  }
+
+  static clearTag008() {
     tag008HoldingsBytesProperties.getAllProperties().forEach(byteProperty => {
       cy.do(QuickMarcEditorRow({ tagValue: '008' }).find(byteProperty.interactor).fillIn(''));
     });
-    pressSaveAndClose();
-    HoldingsRecordView.waitLoading();
+    this.pressSaveAndClose();
```
BogdanDenis, line 225: Same comment here

BogdanDenis (changes requested):

BogdanDenis (approved):

cypress/support/fragments/quickMarcEditor.js
dmitriyfolio: we can, see https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Classes/static

PR #217, opened 2022-02-09
FAT-1195 Filter circulation log by date

MaksimTerekhov:


KyryloBrener (approved):

cypress/support/fragments/circulation-log/searchPane.js
```diff
@@ -12,6 +13,17 @@ export default {
     ]);
   },
 
+  getLastWeekSearchParam() {
```
dmitriyfolio, line 16: may be these methods should be moved to dateTools as common methods

cypress/support/fragments/circulation-log/searchPane.js
```diff
@@ -12,6 +13,17 @@ export default {
...
+    const today = new Date();
+    const lastWeek = new Date(today.getFullYear(), today.getMonth(), today.getDate() - 7);
+    return `${DateTools.padWithZero(lastWeek.getMonth() + 1)}/${DateTools.padWithZero(lastWeek.getDate())}/${lastWeek.getFullYear()}`;
+  },
+
+  getTodaySearchParam() {
```
dmitriyfolio, line 22: may be these methods should be moved to dateTools as common methods

cypress/support/fragments/circulation-log/searchPane.js
```diff
@@ -40,22 +52,35 @@ export default {
     const dateRegEx = /\d{1,2}\/\d{1,2}\/\d{4},\s\d{1,2}:\d{2}\s\w{2}/gm;
 
     return cy.get('#circulation-log-list').then(element => {
```
dmitriyfolio, line 54: I believe  it be implemented through interactor

cypress/support/fragments/circulation-log/searchPane.js
```diff
@@ -40,22 +52,35 @@ export default {
...
-      const resultCount = element.attr('data-total-count');
+      let resultCount = element.attr('data-total-count');
```
dmitriyfolio, line 55: it can be const  and 1 line with element.attr('data-total-count')>29?29:element.attr('data-total-count')

cypress/support/fragments/circulation-log/searchPane.js
```diff
@@ -40,22 +52,35 @@ export default {
...
+      // only 30 records shows on every page
+      if (resultCount > 29) {
+        resultCount = 29;
+      }
 
       // verify every string in result table
       for (let i = 0; i < resultCount; i++) {
         const resultRow = getResultRowByRowNumber(i);
         cy.do([
-          resultRow.userBarcode.perform(el => expect(el.textContent).to.match(/\d/)),
```
dmitriyfolio, line 49: Do you really need " cy.do([resultRow.userBarcode.perform(el => expect(el.textContent).to.match(/\d|/))," instead of cy.expect(resultRow.userBarcode.......)? Check https://frontside.com/interactors/docs/matchers please

cypress/support/fragments/circulation-log/searchPane.js
```diff
@@ -40,22 +52,35 @@ export default {
...
+          resultRow.userBarcode.perform(el => expect(el.textContent).to.match(/\d|/)),
```
dmitriyfolio, line 66: you collect related with row interactors into object. You can add regexps to this object too and apply something like row.cells.foreach(cell => cell.interactor.match(cell.macther))

cypress/integration/circulation-log/search.spec.js
```diff
@@ -26,4 +26,9 @@ describe('navigate to circulation log', () => {
 
     SearchPane.verifyResultCells();
   });
+
+  it('C16976 Filter circulation log by date', { tags: [TestTypes.smoke] }, () => {
+    SearchPane.filterByLastWeek();
+    SearchPane.verifyResultCells();
```
dmitriyfolio, line 32: based on TC I see it's critical to check date values in rows after filtering. If I'm right, need to add such checks into your test

cypress/support/fragments/circulation-log/searchPane.js
MaksimTerekhov: I thought about that, but this method returns value with custom formmating. I didn't see were this formating need except inventory search

cypress/support/fragments/circulation-log/searchPane.js
MaksimTerekhov: I thought about that, but this method returns value with custom formmating. I didn't see were this formating need except inventory search

cypress/support/fragments/circulation-log/searchPane.js
MaksimTerekhov: Unfortunatly nope, because this is container and we don't have interactor for this element. And we don't need interactor for this element because it seems not usual case

cypress/support/fragments/circulation-log/searchPane.js
dmitriyfolio: you can add methods related with circ logic and may be someone find it usable in another app and reuse it and rename it to more common meaning

cypress/support/fragments/circulation-log/searchPane.js
dmitriyfolio: let's add todo related with redesign to interactors and try to make this later.

cypress/support/fragments/circulation-log/searchPane.js
dmitriyfolio: let's move it into method related with circ log search  in dateTools. File already has a few special date frmatters

MaksimTerekhov: Do you think it is good practice to collect special formatter inside general tools that ca be using in whole project?
May be better to use special formatter inside place where it need?

dmitriyfolio (approved):

PR #221, opened 2022-02-10
UITEST-91: Replace Hmac by Hash avoiding hardcoded secret

julianladisch:
snyk.io correctly flags the empty string in crypto.createHmac('sha256', '') as a security issue because the secret of a Hmac must not be a public constant:

https://cwe.mitre.org/data/definitions/547.html

Replacing the Hmac algorithm by a hash algorithm is the way to fix it because only a hash is needed here.


zburke (changes requested):
@julianladisch , it looks like this file is cruft from our long-deprecated NightmareJS tests, but we forgot to remove it when we removed Nightmare in #101. Thanks for pointing out the security issue. Rather than adjusting the code, let's just remove the file completely.

zburke:
Replaced by #254

PR #237, opened 2022-02-24
FAT-1201 Verify "Browse call numbers" option on the Instances tab

MaksimTerekhov:


cypress/integration/inventory/call-number-browse.spec.js
```diff
@@ -0,0 +1,19 @@
...
+
+describe('verify call number browse feature', () => {
+  beforeEach('navigate to inventory', () => {
+    cy.login(Cypress.env('diku_login'), Cypress.env('diku_password'));
+    cy.visit(TopMenu.inventoryPath);
+  });
+
+  it('C347902 Verify "Browse call numbers" option on the Instances tab', { tags: [TestTypes.smoke] }, () => {
+    InventorySearch.verifyKeywordsAsDefault();
+    InventorySearch.selectCallNumberBrowse();
```
dmitriyfolio, line 14: rename to selectBrowseCallNumbers()

cypress/integration/inventory/call-number-browse.spec.js
```diff
@@ -0,0 +1,19 @@
...
+    InventorySearch.verifyCallNumberBrowseEmptyPane();
```
dmitriyfolio, line 15: "The label "Browse inventory" is displayed on the main pane" - can't see it. May be  bug or outdated test

dmitriyfolio: the same related with ""Browse for results entering a query or choosing a filter" on the main pane is displayed."

cypress/support/fragments/inventory/inventorySearch.js
```diff
@@ -47,6 +58,42 @@ export default {
     return cy.do(keywordInput.fillIn(kw));
   },
 
+  selectCallNumberBrowse() {
+    // cypress can't draw selected option without wait
+    cy.wait(1000);
+    return cy.do(Select('Search field index').choose('Browse call numbers'));
```
dmitriyfolio, line 64: why do you add this return statement?

MaksimTerekhov: I can use it with promise like

`selectCallNumberBrowse.then(element => something)`

I think it is not obligatory, but it can be useful

cypress/integration/inventory/call-number-browse.spec.js
MaksimTerekhov: These checks indide funtion verifyCallNumberBrowseEmptyPane:

```
verifyCallNumberBrowseEmptyPane() {
    const callNumberBrowsePane = Pane({ title: 'Browse inventory' });
    cy.expect(callNumberBrowsePane.exists());
    cy.expect(callNumberBrowsePane.has({ subtitle: 'Enter search criteria to start browsing' }));
    cy.expect(HTML(including('Browse for results entering a query or choosing a filter.')).exists());
  },
```

cypress/support/fragments/inventory/inventorySearch.js
dmitriyfolio: I prefer to keep code related with Present. Future is not predictable. If you sure that you will use it - you are right

PR #240, opened 2022-02-28
Fat 1461 stabilization

dmitriyfolio:


cypress/integration/data-import/view-all.spec.js
```diff
@@ -37,7 +37,7 @@ describe('ui-data-import: Search the "View all" log screen', () => {
     });
   });
 
-  it('C11112 Search the "View all" log screen', { tags: [testTypes.smoke] }, () => {
+  it.only('C11112 Search the "View all" log screen', { tags: [testTypes.smoke] }, () => {
```
MaksimTerekhov, line 40: Do we need this change? Seems like debug option

MaksimTerekhov (approved):

PR #248, opened 2022-03-03
FAT-1203 Verify "Browse call numbers" option on Item tab

MaksimTerekhov:


TetianaParanich (approved):

cypress/support/fragments/inventory/inventorySearch.js
```diff
@@ -151,5 +151,13 @@ export default {
   },
   switchToHoldings: () => {
     cy.do(Button({ id: 'segment-navigation-holdings' }).click());
+  },
+
+  instanceTabIsDefault() {
+    cy.do(
+      Button({ id: 'segment-navigation-instances' }).perform(element => {
```
dmitriyfolio, line 158: are you sure that it can't be implemented through including()/matching()? There is example from frontside:
Link({
  href: and(
    matching(/^https/),
    including('google'),
    matching(/\.com$/)
  )
}).exists();

dmitriyfolio (approved):

cypress/support/fragments/inventory/inventorySearch.js
MaksimTerekhov: Default instance has a class "primary". With your suggestions I don't see how I can check class. May be I'm wrong

cypress/integration/inventory/call-number-browse.spec.js
```diff
@@ -22,4 +22,11 @@ describe('verify call number browse feature', () => {
     InventorySearch.verifyKeywordsAsDefault();
     InventorySearch.browseCallNumberIsAbsent();
   });
+
+  it('C347923 Verify "Browse call numbers" option on Item tab', { tags: [TestTypes.smoke] }, () => {
```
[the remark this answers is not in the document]
MaksimTerekhov: We talked about this already, this 

> User is logged into FOLIO with "Inventory:View instances, holdings, and items" permissions

means about minimal permissions that user must have. I leave TODO for future and thinking about creating new user

PR #275, opened 2022-03-18
FAT-817: Implement e-2-e automation of test case C1294

Javokhir12:
## Purpose
- Implement e-2-e automation of test case `C1294`: Create a Holdings record as another user than the one that created the Instance

## Refs
- https://issues.folio.org/browse/FAT-817

### DIFF
--- a/cypress/integration/inventory/create-holdings-as-different-user.spec.js
+++ b/cypress/integration/inventory/create-holdings-as-different-user.spec.js
@@ -6,6 +6,7 @@ import permissions from '../../support/dictionary/permissions';
 import InventoryInstance from '../../support/fragments/inventory/inventoryInstance';
 import HoldingsRecordView from '../../support/fragments/inventory/holdingsRecordView';
 import getRandomPostfix from '../../support/utils/stringTools';
+import testTypes from '../../support/dictionary/testTypes';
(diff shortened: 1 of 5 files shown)


cypress/integration/inventory/create-holdings-as-different-user.spec.js
```diff
@@ -0,0 +1,61 @@
+import TopMenu from '../../support/fragments/topMenu';
+import InventoryInstances from '../../support/fragments/inventory/inventoryInstances';
+import searchInventory from '../../support/fragments/data_import/searchInventory';
+import { MultiColumnListCell } from '../../../interactors';
+import permissions from '../../support/dictionary/permissions';
+import InventoryInstance from '../../support/fragments/inventory/inventoryInstance';
+import HoldingsRecordView from '../../support/fragments/inventory/holdingsRecordView';
+import getRandomPostfix from '../../support/utils/stringTools';
+import testTypes from '../../support/dictionary/testTypes';
```
KyryloBrener, line 9: Minor comment: let's use same approach, capital letters for all imports. e.g. TestTypes instead of testTypes etc.

KyryloBrener (approved):

PR #283, opened 2022-03-22
greening - FAT-1613

MaksimTerekhov:


cypress/integration/circulation-log/search.spec.js
```diff
@@ -7,7 +7,7 @@ import permissions from '../../support/dictionary/permissions';
 const ITEM_BARCODE = `123${getRandomPostfix()}`;
 let userId = '';
 
-describe('ui-circulation-log: filters', () => {
+describe('ui-circulation-log: filters', { retries: 1 }, () => {
```
dmitriyfolio, line 10: have you checked all another options?

MaksimTerekhov: locally test run is green, logs are empty
if you can get suggest other solution I will be happy tested it

dmitriyfolio (approved):

PR #291, opened 2022-03-24
FAT-1193 Filter circulation log on user barcode

MaksimTerekhov:


cypress/support/fragments/circulation-log/searchPane.js
```diff
@@ -11,6 +11,13 @@ export default {
     cy.get('[class^="button-"][type="submit"]').first().click();
   },
 
+  searchByUserBarcode(barcode) {
+    cy.do([
+      TextField({ name: 'userBarcode' }).fillIn(barcode),
+    ]);
+    cy.get('[class^="button-"][type="submit"]').first().click();
```
dmitriyfolio, line 18: why so difficult?

dmitriyfolio (approved):

cypress/support/fragments/circulation-log/searchPane.js
MaksimTerekhov: Because we have 2 identical buttons, but first section without name, id or another selector =(
image

PR #313, opened 2022-04-05
FAT-1641-greening-lotus-branch

IhorBohdan:


cypress.json
```diff
@@ -3,7 +3,7 @@
   "viewportWidth": 1920,
   "viewportHeight": 1080,
   "video": false,
-  "defaultCommandTimeout": 31000,
+  "defaultCommandTimeout": 151000,
```
dmitriyfolio, line 6: should be supported in cypress\support\index.js too, see https://wiki.folio.org/display/DQA/Interactors#Interactors-Timeout

PR #319, opened 2022-04-05
Fat 1639 greening lotus branch

dmitriyfolio:


interactors/spinner.js
```diff
@@ -0,0 +1,4 @@
+import HTML from './baseHTML';
+
+export default HTML.extend('spinner')
+  .selector('div[class^=spinner-]');
```
IhorBohdan, line 4: Maybe locator and filters should be added?

IhorBohdan (approved):

interactors/spinner.js
dmitriyfolio: it's not needed in current code. Code can be extended if it will needed

PR #332, opened 2022-04-07
Fat 1639 greening lotus branch

dmitriyfolio:


cypress/integration/settings/users/owners.spec.js
```diff
@@ -20,7 +20,7 @@ describe('Management of n fee/fine owners and service points', () => {
       // clarify should be service points be shared between existing users
       cy.allure().startStep('Check presented owners and related service points');
       UsersOwners.getUsedServicePoints().then(usedServicePoints => {
-        addedServicePoints.push(UsersOwners.defaultServicePoints.filter(servicePoint => !usedServicePoints?.includes(servicePoint))[0]);
+        addedServicePoints.push(...UsersOwners.defaultServicePoints.filter(servicePoint => !usedServicePoints?.includes(servicePoint)));
         cy.allure().endStep();
```
MaksimTerekhov, line 24: In future will be better if we put allure marks inside fragments instead tests

MaksimTerekhov (approved):

cypress/integration/settings/users/owners.spec.js
dmitriyfolio: xz

PR #333, opened 2022-04-07
FAT-1643 remove todo from export tests

MaksimTerekhov:


cypress/integration/inventory/moving-items.spec.js
```diff
@@ -9,30 +9,95 @@ import HoldingsRecordView from '../../support/fragments/inventory/holdingsRecord
...
+      permissions.inventoryAll.gui,
+      permissions.uiInventoryMoveItems.gui
+    ])
+      .then(userProperties => {
+        userId = userProperties.userId;
+        cy.login(userProperties.username, userProperties.password);
+        cy.visit(TopMenu.inventoryPath);
+        cy.getToken(Cypress.env('diku_login'), Cypress.env('diku_password'))
+          .then(() => {
+            cy.getLoanTypes({ limit: 1 });
```
dmitriyfolio, line 33: move complicated data preparing into appropriate fragment

MaksimTerekhov: It seems excess, because we use separate and understanding methods like getLoanTypes, createInstance etc

MaksimTerekhov: It seems excess because we use simple and understanding methods likegetLoanTypes, createInstance etc
Also I see a whole promise chain in one place, It gets debugging easilier

IhorBohdan (approved):

PR #338, opened 2022-04-11
FAT-1198 Flag a location as remote storage

MaksimTerekhov:


cypress/support/fragments/settingsMenu.js
```diff
@@ -48,6 +48,7 @@ export default {
   remoteStoragePath: 'settings/remote-storage',
   tagsPath: 'settings/tags',
   tenantPath: 'settings/tenant-settings',
+  locationsPath: 'settings/tenant-settings/location-locations',
```
IhorBohdan, line 51: Maybe we can rename this like tenantLocationsPath?

IhorBohdan (approved):

PR #354, opened 2022-04-28
Fat 1623 tc c350667

dmitriyfolio:


cypress/support/fragments/settings/dataImport/dataImportSettingsMatchProfiles.js
```diff
@@ -0,0 +1,78 @@
+import getRandomPostfix from '../../../utils/stringTools';
```
IhorBohdan, line 1: It seems to me that the files dataImportSettingsMatchProfiles.js,dataImportSettingsMappingProfiles.js,dataImportSettingsJobProfiles.js,dataImportSettingsActionProfiles.js can be named without the prefix dataImport.

IhorBohdan (approved):

PR #439, opened 2022-06-16
UITEST-78 - properly implement 'column' filter of MultiColumnList

JohnC-80:
Currently, the 'column' filter of the MCLCell interactor only uses `textContent`
This PR implements the column filter so that assertions can use the column header's text to locate their applicable cell (just as users do.)
Documentation is updated with an example usage.

**This will break existing tests that use the `column` filter improperly - they should use the `content` filter of MCLCellInteractor instead.**

I went ahead and updated the files to use `content` rather than `column` so that the tests within stripes-testing will work as well as placed a PR for the update in stripes-components: [#1830](https://github.com/folio-org/stripes-components/pull/1830)


dmitriyfolio (approved):

TetianaParanich (approved):

kimkate303 (approved):

JohnC-80:
@kimkate303 @dmitriyfolio  @TetianaParanich 
I'm going to split this PR between 3 things -
1) PR for tests to use the interactor properly
2) PR to bump the major version
3) PR for the actual breaking change

penfriend:
Hello, @JohnC-80 ! Can we close this pull request? Do we need this pull request in open status?

JohnC-80:
Closing, will re-add changes to mcl interactor in separate PR

PR #443, opened 2022-06-20
fix for tests C347621 and C347623

kimkate303:


kimkate303:
@dmitriyfolio should i remove tag <broken>? or we need it to run broken tests separately?

dmitriyfolio (approved):

PR #455, opened 2022-06-23
Change service point function

TetianaParanich:


cypress.json
```diff
@@ -16,4 +16,4 @@
     "grepFilterSpecs": true,
     "grepOmitFiltered": true
   }
-}
\ No newline at end of file
+}
```
dmitriyfolio, line 19: do you need this change?

cypress/integration/circulation-log/search.spec.js
```diff
@@ -81,21 +86,23 @@ describe('ui-circulation-log', () => {
   });
 
   after('Delete all data', () => {
-    CheckinActions.createItemCheckinApi({
+    cy.wrap(CheckinActions.createItemCheckinApi({
```
dmitriyfolio, line 89: you don't need wrap in this place

cypress.json
TetianaParanich: ,I deleted it

cypress/integration/circulation-log/search.spec.js
TetianaParanich: I deleted it

IhorBohdan (approved):

PR #540, opened 2022-08-06
FAT-2395 C356808 Verify uploading file with Item former identifiers

MaksimTerekhov:


TetianaParanich (approved):

cypress/integration/bulk-edit/in-app/bulk-edit-in-app-former-ids.spec.js
```diff
@@ -0,0 +1,72 @@
...
+let user;
+const item = {
+  instanceName: `testBulkEdit_${getRandomPostfix()}`,
+  itemBarcode: generateItemBarcode(),
+};
+const formerId = generateItemBarcode();
+const validItemFormerIdsFileName = `validItemFormerIds_${getRandomPostfix()}.csv`;
+
+describe('bulk-edit', () => {
+  describe('in-app approach', () => {
```
penfriend, line 23: Why don't you use just one string in one describe?

MaksimTerekhov: answer is here https://github.com/folio-org/stripes-testing/pull/541#discussion_r940135013

penfriend (approved):

PR #557, opened 2022-08-28
Import by oclc tests

MaksimTerekhov:


cypress/integration/data-import/import-by-oclc.spec.js
```diff
@@ -0,0 +1,50 @@
...
+      .then(instance => {
+        InventoryInstance.deleteInstanceViaApi(instance.id);
+      });
+    Users.deleteViaApi(user.userId);
+  });
+
+  it('C193953 Overlay existing Source = MARC Instance by import of single MARC Bib record from OCLC (Prokopovych)', { tags: [testTypes.smoke] }, () => {
+    InventoryActions.import(oclc);
+    InventoryInstance.viewSource();
+    InventoryViewSource.contains('999\tf f\t‡s');
```
penfriend, line 41: Can you please explain the difference between the test case and the test: Verify that 999 ff $s and $i remain the same  and 999\tf f\t‡s ?  Do we need to update the test case?

MaksimTerekhov: In first case I click view source after import
In second case I click view source after inventory search
And verify need here to be sure that right source was opened

IhorBohdan (approved):

cypress/integration/data-import/import-by-oclc.spec.js
MaksimTerekhov: In test I verified 999 f f but with tabs \t as in test case
I assume you mean difference between $i and $s - it depends of oclc code. In test case author suggest to pick code from third-party site. May be author write test case for one book but I picked another

penfriend (approved):

PR #558, opened 2022-08-30
RANCHER-89 Added cypress-testrail-simple plugin

kdubniak:


cypress/plugins/index.js
```diff
@@ -58,3 +58,7 @@ module.exports = (on, config) => {
   return config;
 };
 
+module.exports = async (on, config) => {
```
MaksimTerekhov, line 61: This will crash our tests)
Because you reassign module.exports from 11 line
Better to add async in 11 line and move your code inside first module.exports block

MaksimTerekhov (approved):

PR #573, opened 2022-09-13
test rail integration copy from master

MaksimTerekhov:


cypress/plugins/index.js
```diff
@@ -55,6 +55,7 @@ module.exports = (on, config) => {
     },
   });
   allureWriter(on, config);
+  await require('cypress-testrail-simple/src/plugin')(on, config);
```
penfriend, line 58: This is not enough. Please add this package to the package.json file

penfriend (approved):

OllIsanbaev (approved):

PR #598, opened 2022-09-30
FAT-1316:C4060 Test the search and filter options for fund groups

IhorBohdan:


MaksimTerekhov (approved):

cypress/integration/finance/groups/search-and-gilter.spec.js
```diff
@@ -0,0 +1,79 @@
...
+
+            Groups.createViaApi(defaultGroup)
+              .then(groupResponse => {
+                defaultGroup.id = groupResponse.id;
+              });
+          });
+      });
+    cy.createTempUser([
+      permissions.uiFinanceCreateViewEditGroups.gui,
+      permissions.uiFinanceViewEditDeletGroups.gui,
```
penfriend, line 36: Are you sure that you need these two permissions?
According to the test-case the user should have permission to view groups. There is a permission with name "Finance: View group". Please clarify with PO, manual tester if your permission is valid in this case

penfriend (changes requested):

cypress/integration/finance/groups/search-and-gilter.spec.js
IhorBohdan: Deleted

TetianaParanich (approved):

penfriend (approved):

PR #608, opened 2022-10-11
FAT-3079/C360521

OllIsanbaev:
https://foliotest.testrail.io/index.php?/cases/view/360521
https://issues.folio.org/browse/FAT-3079


cypress/support/fragments/data_import/logs/logs.js
```diff
@@ -50,4 +51,24 @@ export default {
...
+        .find(Link(title)).perform(element => {
+          cy.visit(element.href);
+        })
+    );
+  },
+
+  checkAuthorityLogJSON: () => {
+    cy.do(Button('Authority').click());
+    cy.expect([
+      HTML('"sourceFileId":').exists(),
```
MaksimTerekhov, line 68: Don't use naked HTML pls
It will make a lot of difficulties in debugging
Try to use HTML after another any selector like this

TestField().find(HTML())

MaksimTerekhov (approved):

cypress/support/fragments/data_import/logs/logs.js
OllIsanbaev: OK, I understand. But in this case, I had to use only HTML, since the page is a textual representation of JSON.
```
<code class="info---bee3O">{
  <span class="key---wzTh6">"id":</span> <span class="string---LFT5x">"d47173e9-d5e8-40bd-b785-e89d34b5342c"</span>,
  <span class="key---wzTh6">"snapshotId":</span> <span class="string---LFT5x">"1cbfbaaf-0980-4ed3-9eaa-1cb6492d2cf8"</span>,
```

cypress/e2e/marc-authority/marc-authority-import.cy.js
```diff
@@ -0,0 +1,48 @@
...
+    });
+  });
+
+  after('Deleting data', () => {
+    Users.deleteViaApi(testData.userProperties.userId);
+
+    cy.loginAsAdmin({ path: TopMenu.dataImportPath, waiter: DataImport.waitLoading });
+    DataImport.selectLog();
+    DataImport.openDeleteImportLogsModal();
+    DataImport.confirmDeleteImportLogs();
```
penfriend, line 35: I have not found the deletion of three marc authorities that are created after the execution of the autotest.
I am talking about mark authorities with the names:
1)Apple & Honey Productions
2)Association pour la promotion et la protection de la liberté d'expression au Burundi
3)Apple Academic Press
after the execution of autotest you can found them by  this path: /marc-authorities/authorities/
you can delete it 
to remote one record you can send the api request with method DELETE to the path: records-editor/records/cde37939-fd36-47c8-a5e0-17d873b2c0d7
For example you can found the recordId in the response of the GET api request to the path: 
/search/authorities?limit=100&offset=0&query=%28keyword%3D%3D%22Apple%20Academic%22%29

penfriend (changes requested):

penfriend (approved):

PR #712, opened 2022-12-09
C359237 Verify "Expiration date" option in the dropdown

MaksimTerekhov:
https://foliotest.testrail.io/index.php?/cases/view/359237


re-jas (approved):

IhorBohdan (approved):

JohnC-80:
This PR breaks tests in stripes-components when the calendar is open. The expanded selector causes an error:

```
AmbiguousElementError: datepicker matches multiple elements:
        
        - <div class="container---2BEIC" data-test-datepicker-container="true">
        - <div role="application" aria-roledescription="datepicker" aria-label="calendar" class="calendar---3Cw1O maxWidth---TdFVf" id="datepicker-calendar-container-dp-750" tabindex="-1" data-test-calendar="true">
```

if the interactor is used without any further filtering. If the calendar needs to be checked, it has its own interactor that can be used.

PR #715, opened 2022-12-12
Revert "C359237 Verify "Expiration date" option in the dropdown"

JohnC-80:
Reverts folio-org/stripes-testing#712

Changes to the datepicker interactor break existing tests within stripes components. The expanded selector is causing the following error when the Calendar of the datepicker is open:

```
AmbiguousElementError: datepicker matches multiple elements:
        
        - <div class="container---2BEIC" data-test-datepicker-container="true">
        - <div role="application" aria-roledescription="datepicker" aria-label="calendar" class="calendar---3Cw1O maxWidth---TdFVf" id="datepicker-calendar-container-dp-750" tabindex="-1" data-test-calendar="true">
```

The Calendar pop-up from Datepicker has its own `Calendar` interactor that can be used to check for the existence of the Calendar, if necessary.


MaksimTerekhov (approved):

JohnC-80:
Closed since removing the extra selector in #716 was a better approach.

PR #721, opened 2022-12-19
C359585 Verify clicking on the "Commit changes" button

MaksimTerekhov:
https://foliotest.testrail.io/index.php?/cases/view/359585

### DIFF
--- a/cypress/e2e/bulk-edit/in-app/bulk-edit-in-app-user-uuids.cy.js
+++ b/cypress/e2e/bulk-edit/in-app/bulk-edit-in-app-user-uuids.cy.js
@@ -138,7 +138,7 @@ describe('bulk-edit', () => {
       BulkEditActions.verifyCalendarItem();
     });
 
-    it.only('C359585 Verify clicking on the "Commit changes" button (firebird)', { tags: [testTypes.smoke, devTeams.firebird] }, () => {
+    it('C359585 Verify clicking on the "Commit changes" button (firebird)', { tags: [testTypes.smoke, devTeams.firebird] }, () => {
       BulkEditSearchPane.selectRecordIdentifier('User UUIDs');
 
       BulkEditSearchPane.uploadFile(userUUIDsFileName);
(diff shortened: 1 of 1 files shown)


OllIsanbaev (approved):

cypress/e2e/bulk-edit/in-app/bulk-edit-in-app-user-uuids.cy.js
```diff
@@ -137,5 +137,24 @@ describe('bulk-edit', () => {
       BulkEditActions.openInAppStartBulkEditFrom();
       BulkEditActions.verifyCalendarItem();
     });
+
+    it('C359585 Verify clicking on the "Commit changes" button (firebird)', { tags: [testTypes.smoke, devTeams.firebird] }, () => {
```
penfriend, line 141: It looks like that there is no need in permission: "Users: Can view user profile". So we can remove this permission from the test case

cypress/e2e/bulk-edit/in-app/bulk-edit-in-app-user-uuids.cy.js
```diff
@@ -137,5 +137,24 @@ describe('bulk-edit', () => {
...
+      BulkEditActions.verifyCheckedDropdownMenuItem();
+      BulkEditActions.verifyUncheckedDropdownMenuItem();
+
+      BulkEditActions.openInAppStartBulkEditFrom();
+      BulkEditActions.fillPatronGroup('graduate (Graduate Student)');
+      BulkEditActions.confirmChanges();
+      BulkEditActions.verifyAreYouSureForm(1, user.username);
+      BulkEditActions.commitChanges();
+      BulkEditSearchPane.waitFileUploading();
+      BulkEditActions.verifySuccessBanner(1);
```
penfriend, line 157: It is better to check if the value is been updated also. There is a note about it in the expected result column on the step 8 in the test case: The Preview of record changed is shown with only updated values

penfriend (approved):

cypress/e2e/bulk-edit/in-app/bulk-edit-in-app-user-uuids.cy.js
MaksimTerekhov: done

cypress/e2e/bulk-edit/in-app/bulk-edit-in-app-user-uuids.cy.js
MaksimTerekhov: done

PR #754, opened 2023-01-09
fixed tests

MaksimTerekhov:


cypress/support/fragments/inventory/inventoryInstance.js
```diff
@@ -254,6 +251,11 @@ export default {
       moveItemsButton.click()
     ]);
   },
+
+  closeItemView() {
```
TetianaParanich, line 255: We have the same function in 'support\fragments\inventory\inventoryItem\itemView.js'

TetianaParanich: closeDetailView();

TetianaParanich (approved):

cypress/support/fragments/inventory/inventoryInstance.js
MaksimTerekhov: thx, fixed

PR #830, opened 2023-02-02
fixed a test

nayimovag:
fixed [T1226049](https://foliotest.testrail.io/index.php?/tests/view/1226049)

### DIFF
--- a/cypress/support/fragments/requests/newRequest.js
+++ b/cypress/support/fragments/requests/newRequest.js
@@ -10,6 +10,7 @@ const enterItemBarcodeButton = Button({ id: 'clickable-select-item' });
 const enterRequesterBarcodeButton = Button({ id: 'clickable-select-requester' });
 const saveAndCloseButton = Button('Save & close');
 const selectServicePoint = Select({ name:'pickupServicePointId' });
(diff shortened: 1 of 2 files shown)


MaksimTerekhov (approved):

TetianaParanich (approved):

cypress/support/fragments/requests/newRequest.js
```diff
@@ -32,6 +33,7 @@ export default {
     cy.do(itemBarcodeInput.fillIn(newRequest.itemBarcode));
     cy.intercept('/circulation/loans?*').as('getLoans');
     cy.do(enterItemBarcodeButton.click());
+    cy.do(selectRequestType.choose(newRequest.requestType));
```
re-jas, line 36: this method is used not only in this test so this change could crash test where newRequest.requestType is not specified. 
if you want to keep it than please go to hold-slip.cy.js and also add "requestType" property
image

re-jas (changes requested):

cypress/support/fragments/requests/newRequest.js
nayimovag: fixed

re-jas (approved):

PR #887, opened 2023-02-22
UITEST-106: update MultiColumnList interactor actions for pagination buttons

JohnC-80:
MultiColumnList's pagination was previously rendered within the element's primary container. This caused accessibility tests to report issues.
[The PR in stripes components](https://github.com/folio-org/stripes-components/pull/1993) moves the pagination outside of the container. This PR updates those actions to cover the case. It's possible, with such specific actions, that the `Button` interactor may just have to be used for tests instead.


MaksimTerekhov (approved):

IhorBohdan (approved):

penfriend (approved):

JohnC-80:
Closing this, as it's unnecessary with the implementation of https://github.com/folio-org/stripes-components/pull/2125

PR #926, opened 2023-03-07
UITEST-107: Dockerfile: apt-get upgrade

julianladisch:
Add "apt-get upgrade" to Dockerfile to upgrade Ubuntu packages to available patch versions.

This is needed for software supply chain security and is a security best practice:

https://github.com/folio-org/folio-tools/tree/master/folio-java-docker/openjdk17#sample-module-dockerfile https://dev.folio.org/guides/best-practices-dockerfiles/


zburke (commented):
No objections from me as this looks syntactically valid, but I don't feel qualified to approve this PR. 

Is this how we do things in CI? If so, can/should we add comments here that we need to keep `.github/workflows/*` and `Dockerfile` in sync? If this isn't what we do in CI, is there a reason to do things differently here? In an ideal world, I would hope those envs would be closely aligned if not identical.

zburke (approved):

PR #1058, opened 2023-04-05
fixed C15850

TetianaParanich:


OllIsanbaev (approved):

cypress/support/fragments/settings/inventory/instance-holdings-item/fastAdd.js
```diff
@@ -1,26 +1,20 @@
...
-            `{"instanceStatusCode":"${value}","defaultDiscoverySuppress":"true"}`
-          },
-          isDefaultSearchParamsRequired: false,
-        });
-      });
+export default {
+  calloutMessages,
+  changeDefaultInstanceStatus:(statusCode) => {
+    cy.do(Select({ name:'instanceStatusCode' }).choose(statusCode));
+    cy.get('button[type="submit"]').then(element => {
```
MaksimTerekhov, line 13: use comment here like this example:

// Interactor Button are not allow to pick both buttons with disable attr and without disabled attr

MaksimTerekhov (approved):

PR #1383, opened 2023-07-19
eholdingsExports/pavithraideyalabs

pavithradasari:
C356417,C367972,C353217


cypress.config.js
```diff
@@ -81,6 +83,6 @@ module.exports = defineConfig({
 
       return config;
     },
-    baseUrl: 'https://folio-testing-cypress-diku.ci.folio.org',
+    baseUrl: 'https://bugfest-orchid.int.aws.folio.org/',
```
ostapwd, line 86: what is the reason for switching to this env?

ostapwd (changes requested):

ostapwd (approved):

PR #1406, opened 2023-07-22
Check out/santosh ideyalabs

Vineethasakhamuri:


cypress/support/commands.js
```diff
@@ -1 +1,9 @@
 require('cypress-downloadfile/lib/downloadFileCommand');
+
+
+
+Cypress.on('uncaught:exception', (err, runnable) => {
```
ryandberger, line 5: Let's add `console.log(err?.message)` here so we at least see the exception in the console.

cypress/support/fragments/users/usersSearchPane.js
```diff
@@ -38,11 +61,92 @@ export default {
...
+    cy.do([ActionButton.click(),
+      editButton.click(),
+      Select(name).choose(label),
+      saveAndClose.click(),
+      AdditionalInfo.click()
+    ]);
+  },
+
+  DragAndDropCustomFields: () => {
+    cy.get('.FieldAccordionDraggableWrapper---c2vRL').then((elements) => {
```
ryandberger, line 131: Should be `[class^=FieldAccordionDraggableWrapper---]`

cypress/support/commands.js
Vineethasakhamuri: Okay @ryandberger

PR #1681, opened 2023-09-27
C581 - verify change load due date

tomleexx:
add new scripts for C581 "Loan Details: Test change due date (TaaS)"


tomleexx:
will create another pull request from this repositories

PR #1706, opened 2023-09-29
Fat 8720

ostapwd:


TetianaParanich (approved):

JohnC-80:
@ostapwd The change from this PR breaks the implementation of the multi-column-list interactor. if the selector is set to `[role=grid]`, the actions to `clickNextPagingButton` and `clickPreviousPagingButton`  will no longer work.

PR #2039, opened 2023-10-25
C380427 A new MARC bib record is opened in QuickMARC UI (spitfire) (TaaS)

elen2345:
[testRail](https://foliotest.testrail.io/index.php?/cases/view/380427)
[AllureReport](https://jenkins-aws.indexdata.com/job/folioRancher/job/folioTestingTools/job/runCypressTests/5079/allure)
image

### DIFF
--- a/cypress/support/fragments/quickMarcEditor.js
+++ b/cypress/support/fragments/quickMarcEditor.js
@@ -261,6 +261,7 @@ defaultFieldValues.getSourceContent = (contentInQuickMarcEditor) => contentInQui
 const requiredRowsTags = ['LDR', '001', '005', '008', '999'];
 const readOnlyAuthorityTags = ['LDR', '001', '005', '999'];
 const readOnlyHoldingsTags = ['001', '004', '005', '999'];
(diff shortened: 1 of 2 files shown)


cypress/support/fragments/quickMarcEditor.js
```diff
@@ -1531,4 +1532,14 @@ export default {
...
   },
+
+  checkDefaultContent() {
+    this.checkContent('00000n\\\\\\a2200000uu\\4500', 0);
+    this.checkFieldsExist(expectedFields);
+    this.checkEmptyContent('001');
+    this.checkEmptyContent('005');
+    this.checkEmptyContent('999');
+    this.checkEmptyContent('008');
+    this.verifyTagField(4, '245', '\\', '\\', '$a ', '');
```
zentestuken, line 1543: Could you please also add a check with `checkContent()` method for 245 field to make sure that "$a " is the only thing that is shown there (`verifyTagField()` uses `including` matcher, so it will pass even if there is some value after $a or even other subfields)?

dnovikau (approved):

PR #2076, opened 2023-10-27
C411787 Check that user can create TLR Hold for Instance with no Items (vega) (TaaS)

IlyaLyskavets:
[Test Rails](https://foliotest.testrail.io/index.php?/cases/view/411787)

Allure 
image


cypress/support/fragments/settings/circulation/titleLevelRequests.js
```diff
@@ -72,6 +75,28 @@ export default {
       });
   },
 
+  changeFailToCreateHoldForBlockedRequest(status) {
```
manowarqqq, line 78: method doesnt work if status 'allow'
method doesn't work if checkbox already selected

manowarqqq (changes requested):

cypress/support/fragments/settings/circulation/titleLevelRequests.js
IlyaLyskavets: used suggested approach
and added 
uncheckIfSelected() to the checkbox interactor since it needs to be unchecked

manowarqqq (approved):

manowarqqq:
For the future tests please use git amend to have test in one commit

ostapwd (approved):

PR #2644, opened 2023-12-08
C356848 Verify that "Source" value displays only the Last name of user, which edited record, when First name of user is not populated (spitfire)

manowarqqq:
https://foliotest.testrail.io/index.php?/cases/view/356848


cypress/e2e/marc/marc-authority/edit-authority-record/verify-source-value-displays-only-last-name.cy.js
```diff
@@ -0,0 +1,111 @@
...
+        MarcAuthority.waitLoading();
+
+        // Step 8: Click on the "Actions" button and select the "Edit" option
+        MarcAuthority.edit();
+        QuickMarcEditor.waitLoading();
+
+        // Step 9: Verify that "Source" value matches with your user's Last name
+        QuickMarcEditor.checkUserNameInHeader(
+          testData.userProperties.firstName,
+          testData.userProperties.lastName,
```
zentestuken, line 106: Could you please update the test so that:
- test user would only have Last name and no First name specified (probably would require additional/updated methods for user creation)
- check that only Last name of the user is then shown in header
This check is described in the test case and is the main idea of it

zentestuken, line 106: Could you please update the test so that:
- test user would only have Last name and no First name specified (probably would require additional/updated methods for user creation)
- check that only Last name of the user is then shown in header
This check is described in the test case and is the main idea of it

zentestuken (changes requested):

zentestuken (approved):

Scorpibear (approved):

PR #2759, opened 2023-12-18
eHoldings - removed user assignments

zentestuken:


ostapwd (approved):

cypress/e2e/eholdings/package/eholdings-export-package-with-10k-plus.cy.js
```diff
@@ -35,7 +34,6 @@ describe('eHoldings', () => {
         Permissions.exportManagerAll.gui,
       ]).then((userProperties) => {
         testData.user = userProperties;
-        AssignedUsers.assignUserToDefaultCredentialsViaApi({ userId: testData.user.userId });
```
dnovikau, line 38: User should be assigned to have access to eHoldings app. I propose not to remove assignement action and add "unassignUserFromDefaultCredentialsViaApi" to after hook. What do you think?

zentestuken: We don't need to assign users when only one KB is configured

dnovikau (approved):

TetianaParanich (approved):

PR #3698, opened 2024-06-04
Eureka capabilities test

PavloPak:


cypress/support/api/users.js
```diff
@@ -141,6 +145,54 @@ Cypress.Commands.add('createTempUser', (permissions = [], patronGroupName, userT
               } else cy.updateRolesForUserApi(userProperties.userId, [roleId]);
             });
           } else {
```
zentestuken, line 147: I think it would be better to add a condition `if (Cypress.env('eureka'))` so that capability assignment would only work on Eureka. That way it would be easier to merge this branch to master later on.

cypress/support/api/users.js
```diff
@@ -141,6 +145,54 @@ Cypress.Commands.add('createTempUser', (permissions = [], patronGroupName, userT
...
+
+                if (capabilitySetsIds.length === 0) {
+                  cy.log('Capability sets not found ');
+                } else {
+                  cy.addCapabilitySetsToNewUserApi(userProperties.userId, capabilitySetsIds);
+                }
+              });
+            });
+            cy.wait(3000);
             cy.addPermissionsToNewUserApi({
```
zentestuken, line 196: For the same reason as in first comment, it would be better to leave the original method logic with permissions in `else` blick.
Also, it looks like `cy.addPermissionsToNewUserApi()` will run even on Eureka as it is now, which is not necessary

zentestuken (approved):

PR #3878, opened 2024-07-05
FAT-14783 Fix multi-select locator

JohnC-80:
A working locator would allow the interactor to be used with the component's visible label rather than rely on filters for potentially generate strings.


sviatlana-stsiapanava (commented):
@JohnC-80  I've reviewed the locator update.
 
 It seems to function well on pages where there's only a single element with the attribute [role=searchbox]. However, for instance in the MARC Authority application, this approach doesn't work because there are three elements identified by this selector and the locator [role=searchbox]. And if I understand correctly, then if there are multiple elements in the DOM that match the selector and locator, the interactor will only consider the first one it finds.
 
  For these pages, I can continue using the Multiselect({ label: 'Type of heading' }) strategy to interact with multiselect elements. But I also noticed that there is an approach where we pass the 'label' into the 'locator', as seen in the Dropdown interactor, for example. If this is a suitable method for writing and using interactors, might it be applicable in Multiselect too? If not, let's stick with the current updates.

JohnC-80:
@sviatlana-stsiapanava in my understanding, the role of the `.locator` within the interactor is to further filter the possible elements when a string is provided to the constructor. So after evaluating what it finds via the `selector` alone, it would further compare the provided string parameter against the result of the `locator` function - providing a way to select via visible label. Granted this **is** also what the `label` filter does, except with the limitation that `label` is only set up for a `label` rendered by the prop... so if the developer has use `aria-labelledby` this approach will not work. If the behaviour that you see is that it *only works on the first element that the `selector` finds, then this seems like a problem with interactors in general.

sviatlana-stsiapanava (approved):

PR #4708, opened 2025-02-07
added check for data import

eremv:


scripts/spitfire.prepare.js
```diff
@@ -182,7 +196,7 @@ async function removeAutotestInstances() {
 
 async function removeAutotestAuthorities() {
   const autotestAuthorities = await axios({
-    url: '/search/authorities?limit=500&query=(keyword="autotest" or keyword="auto")',
+    url: '/search/authorities?limit=500&query=(keyword="autotest" or keyword="auto" or keyword="C*")',
```
zentestuken, line 199: We might also add ` and (authRefType==("Authorized"))` to the query so that it would ignore reference records which essentially refer to the same record as authorized one.
And the same in `removeAuthoritiesWithSpecificSource` function.

scripts/spitfire.prepare.js
```diff
@@ -23,12 +25,24 @@ async function getToken() {
   });
   const setCookieHeader = response.headers['set-cookie'];
   axios.defaults.headers.common.Cookie = setCookieHeader;
+  axios.defaults.headers.common['Content-Type'] = 'application/json';
```
zentestuken, line 28: Would this not conflict with line 85?
`headers: { 'Content-type': 'application/vnd.api+json' },`

scripts/spitfire.prepare.js
```diff
@@ -237,22 +251,209 @@ async function removeUserAuthoritySourceFiles() {
...
+      `${envData.apiBaseUrl}/change-manager/jobExecutions/${jobExecutionId}`,
+    );
+    checkResponse(
+      response,
+      `\t\tGetting parent job execution status (${attempts} attempts left)`,
+      200,
+    );
+    const status = response.data;
+    if (status.status === 'COMMITTED' && status.uiStatus === 'RUNNING_COMPLETE') return status;
+    await new Promise((res) => setTimeout(res, 5_000));
```
zentestuken, line 346: You can use existing `wait()` function here

scripts/spitfire.prepare.js
eremv: In line 85 default Content-Type header will be overriden  by specific 'application/vnd.api+json' so no conflicts expected.

TetianaParanich (approved):

zentestuken (approved):

PR #4927, opened 2025-04-07
UITest-133 Export AppList interactor

JohnC-80:
This PR exports the Applist interactor so that it can be used by other test suites (stripes-core, stripes-smart-components)


zburke (approved):
On the one hand, I HATE default export. On the other, it is a well-established pattern here. Is there a reason to buck the trend?

ostapwd (approved):

PR #5054, opened 2025-05-21
FAT-20132 - update pane header label of create quickmarc pane

valerypilko:
https://folio-org.atlassian.net/browse/FAT-20132

Label of create pane in "quickmarc" was updated.

Allure reports:
Non-ECS:
https://jenkins.ci.folio.org/job/folioTestingTools/job/runCypressTests/8300/allure/#suites/a64df16dd423bcf8901a28ccfb6cab2d/cd8f5d15bf793e1b/

ESC:
https://jenkins.ci.folio.org/job/folioTestingTools/job/runCypressTests/8306/allure/#suites

### DIFF
--- a/cypress/e2e/marc/marc-bibliographic/create-new-marc-bib/create-new-marc-bib-check-pane.cy.js
+++ b/cypress/e2e/marc/marc-bibliographic/create-new-marc-bib/create-new-marc-bib-check-pane.cy.js
@@ -9,7 +9,7 @@ describe('MARC', () => {
   describe('MARC Bibliographic', () => {
     describe('Create new MARC bib', () => {
       const testData = {
-        headerText: 'New MARC bib record',
+        headerText: /New .*MARC bib record/,
       };
 
       before('Create test data', () => {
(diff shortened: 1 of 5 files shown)


cypress/e2e/marc/marc-bibliographic/create-new-marc-bib/create-new-marc-bib-check-pane.cy.js
```diff
@@ -8,7 +8,9 @@ import Users from '../../../../support/fragments/users/users';
 describe('MARC', () => {
   describe('MARC Bibliographic', () => {
     describe('Create new MARC bib', () => {
-      const testData = {};
+      const testData = {
+        headerText: 'New MARC bib record',
```
zentestuken, line 12: I recommend to use regExp here: `headerText: /New .*MARC bib record/,`. That way the test would not fail when we run non-ECS tests on ECS member tenant (which we sometimes do). In this case the header text will be "New local MARC bib record". `checkPaneheaderContains()` method supports regExp as an argument.

zentestuken (approved):

cypress/e2e/marc/marc-bibliographic/create-new-marc-bib/create-new-marc-bib-check-pane.cy.js
valerypilko: Thanks, fixed

PR #5203, opened 2025-06-26
Fix C405530

victorymarkus:
Знімок екрана 2025-06-26 о 14 31 50

### DIFF
--- a/cypress/e2e/inventory/call-number-browse/browse-call-numbers-only-correct-types.cy.js
+++ b/cypress/e2e/inventory/call-number-browse/browse-call-numbers-only-correct-types.cy.js
@@ -131,13 +131,11 @@ describe('Inventory', () => {
     }
 
     function searchAndCheckCallNumbers(type, searchValue) {
-      InventorySearchAndFilter.browseSearch(searchValue);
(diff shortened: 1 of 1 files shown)


yanarusakova (approved):

cypress/e2e/inventory/call-number-browse/browse-call-numbers-only-correct-types.cy.js
```diff
@@ -130,6 +130,17 @@ describe('Inventory', () => {
         .flat();
     }
 
+    function searchAndCheckCallNumbers(type, searchValue) {
```
eremv, line 133: The key moment here is to waitForCallNumberToAppear(callNumber) before performing browseSearch.  Otherwise we risk to not find target value in the results list.  If you wish to group these actions into separate function I would suggest the next implementation:

```

      filterCNOfType(type).forEach((callNumber) => {
        BrowseCallNumber.waitForCallNumberToAppear(callNumber);
        InventorySearchAndFilter.browseSearch(searchValue);
        BrowseCallNumber.valueInResultTableIsHighlighted(searchValue);
        BrowseCallNumber.checkValuePresentInResults(callNumber);
      });

eremv (changes requested):

PR #5468, opened 2025-08-13
Disable broken tests and fix

victorymarkus:
Screenshot 2025-08-13 121124
Screenshot 2025-08-13 111325

### DIFF
--- a/cypress/e2e/marc/marc-authority/marc-authority-search.cy.js
+++ b/cypress/e2e/marc/marc-authority/marc-authority-search.cy.js
@@ -102,7 +102,9 @@ describe('MARC', () => {
       Users.deleteViaApi(testData.userProperties.userId);
       Users.deleteViaApi(testData.userPropertiesC409449.userId);
       createdRecordIDs.forEach((id) => {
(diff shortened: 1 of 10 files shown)


cypress/e2e/marc/marc-authority/marc-authority-search.cy.js
```diff
@@ -102,7 +102,9 @@ describe('MARC', () => {
       Users.deleteViaApi(testData.userProperties.userId);
       Users.deleteViaApi(testData.userPropertiesC409449.userId);
       createdRecordIDs.forEach((id) => {
-        MarcAuthority.deleteViaAPI(id);
+        if (id) {
+          MarcAuthority.deleteViaAPI(id);
```
zentestuken, line 106: I think it's better to call it like that: `MarcAuthority.deleteViaAPI(id, true);`. That way it will ignore a failure (when record not created or when it cannot be deleted, etc.)

zentestuken (approved):

yanarusakova (approved):

PR #5632, opened 2025-09-09
Update MARC tests and API interactions for improved validation and er…

eremv:
…ror handling


sviatlana-stsiapanava (approved):

yanarusakova (approved):

cypress/support/fragments/marcAuthority/marcAuthorities.js
```diff
@@ -1704,11 +1704,13 @@ export default {
 
   deleteMarcAuthorityByTitleViaAPI(title, authRefType = 'Authorized') {
     this.getMarcAuthoritiesViaApi({ limit: 100, query: `keyword="${title}"` }).then((records) => {
-      records.forEach((record) => {
-        if (record.authRefType === authRefType) {
-          this.deleteViaAPI(record.id, true);
-        }
-      });
+      if (records && records.length > 0) {
```
zentestuken, line 1707: Is this condition necessary?
I have updated getMarcAuthoritiesViaApi() to return an empty array if nothing found (which is necessary in itself for tests which use it directly). So, it will return an array in any case but network error.

zentestuken (approved):

PR #5633, opened 2025-09-09
Fat 20589 fixes nightrun

yanarusakova:


cypress/e2e/inventory/tags/find-tag-with-special-characters-using-api.cy.js
```diff
@@ -103,6 +103,7 @@ describe('Inventory', () => {
       'C380422 Find Tag with special characters using API (volaris)',
       { tags: ['extendedPath', 'volaris', 'C380422', 'eurekaPhase1'] },
       () => {
+        cy.waitForAuthRefresh(() => {}, 20_000);
```
eremv, line 106: Why don't you add wait for auth refresh right after login action in before hook?

```
          cy.waitForAuthRefresh(() => {
            cy.login(testData.user.username, testData.user.password);
          });
```

cypress/support/fragments/inventory/inventorySearchAndFilter.js
```diff
@@ -792,6 +792,7 @@ export default {
 
   resetAllAndVerifyNoResultsAppear() {
     cy.do(resetAllButton.click());
+    cy.reload();
```
ostapwd, line 795: auth refresh should be handled here

ostapwd (changes requested):

ostapwd (approved):

eremv (approved):

PR #5634, opened 2025-09-10
fix failed tests

victorymarkus:


cypress/e2e/marc/marc-holdings/moving-items.cy.js
```diff
@@ -100,25 +100,29 @@ describe('MARC', () => {
...
-      cy.getInstance({
-        limit: 1,
-        expandAll: true,
-        query: `"items.barcode"=="${ITEM_BARCODE}"`,
-      }).then((instance) => {
-        cy.deleteItemViaApi(instance.items[0].id);
-        cy.deleteHoldingRecordViaApi(instance.holdings[0].id);
-        cy.deleteHoldingRecordViaApi(instance.holdings[1].id);
-        InventoryInstance.deleteInstanceViaApi(instance.id);
+      cy.waitForAuthRefresh(() => {
```
ostapwd, line 115: why do we need that here?

ostapwd (changes requested):

cypress/e2e/marc/marc-holdings/moving-items.cy.js
victorymarkus: Yesterday, the auth refresh appeared in the after block each time the test ran. Now the problem is gone, so I’ll delete it.

ostapwd (approved):

PR #6906, opened 2026-04-16
FAT-25255-fix-QG

yanarusakova:


cypress/support/fragments/linked-data/work.js
```diff
@@ -41,4 +41,13 @@ export default {
...
   },
+
+  deleteInventoryInstanceViaApi(inventoryId) {
+    return cy.okapiRequest({
+      method: 'DELETE',
+      path: `instance-storage/instances/${inventoryId}`,
+      isDefaultSearchParamsRequired: false,
+      failOnStatusCode: false,
+    });
+  },
```
sviatlana-stsiapanava: We have implemented the same method for instance deletion in the cypress\support\fragments\inventory\inventoryInstance.js fragment:
``

deleteInstanceViaApi: (id) => {
    cy.okapiRequest({
      method: REQUEST_METHOD.DELETE,
      path: `instance-storage/instances/${id}`,
      isDefaultSearchParamsRequired: false,
      failOnStatusCode: false,
    });
  },

``

 I think we could also use it in linked-data tests for consistency

sviatlana-stsiapanava (approved):

cypress/support/fragments/linked-data/work.js
yanarusakova: fixed

PR #7035, opened 2026-05-04
FAT-24344 Spitfire - Create Cypress tests - [Consortia] Verify that user must have "edit" permissions to edit "Display settings" on current tenant

BogdanDenis:
## Description
Spitfire - Create Cypress tests - [Consortia] Verify that user must have "edit" permissions to edit "Display settings" on current tenant
image


BogdanDenis:
Implementation was already in progress in https://folio-org.atlassian.net/browse/FAT-25520

PR #7160, opened 2026-05-20
[FAT-24139] Add Cypress test to verify case-insensitive value filtering in query builder

bvsharp:
[FAT-24139](https://folio-org.atlassian.net/browse/FAT-24139) Add Cypress test to verify case-insensitive value filtering in query builder

### DIFF
--- a/cypress/e2e/lists/query-builder/search-query-builder-api.cy.js
+++ b/cypress/e2e/lists/query-builder/search-query-builder-api.cy.js
@@ -50,7 +50,6 @@ describe('Lists', () => {
       });
     };
 
-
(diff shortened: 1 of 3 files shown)


cypress/e2e/lists/query-builder/search-query-builder-api.cy.js
```diff
@@ -50,7 +50,6 @@ describe('Lists', () => {
       });
     };
 
-
```
bvsharp, line 53: This file is 100% formatting changes

mweaver-ebsco (approved):

PR #7196, opened 2026-05-29
FAT-24443 Spitfire - Create Cypress tests - [Consortia] Link ""Subjects"" fields of Shared record automatically on Member tenant when different subfields selected as linkable in configs of Central and Member tenants

BogdanDenis:
image

[FAT-24443](https://folio-org.atlassian.net/browse/FAT-24443)


zentestuken:
Please make bib titles and authority headers a bit more unique:
1) By switching to dynamic creation of all records via API with unique values
2) OR by simply adding `AT_CXXXXX_<record type>` prefix to each heading/title
Otherwise multiple records with same/similar titles/headers will cause problems in test runs, especially during bugfests and for search/browse tests.
Note that authority naturalIds should also be unique.

eremv (approved):

PR #7526, opened 2026-08-13
FAT-28188: Implemented C667566

Daria-Reznikova:
https://folio-org.atlassian.net/browse/FAT-28188
https://foliotest.testrail.io/index.php?/cases/view/667566

image


cypress/support/constants/orders/order.js
```diff
@@ -117,8 +117,15 @@ export const ORDER_EXPORT_CSV_FIELDS = {
...
   PO_NUMBER: 'PO number',
+  VENDOR: 'Vendor',
+  ORDER_TYPE: 'Order type',
 };
 
-export const ORDER_CALLOUT_MESSAGES = {
-  ORDER_UNOPENED: (orderNumber) => `The Purchase order - ${orderNumber} has been successfully unopened`,
+export const ORDER_AND_ORDER_LINE_BUTTONS = {
+  ADD_POL: 'Add POL',
+  CANCEL: 'Cancel',
```
usavkov-epam: I think "Cancel", "Save & close", "Save & keep editing" are commonly used. I do not think we need to duplicate them https://github.com/folio-org/stripes-testing/blob/38b15c3ca85dffcf780d976a9d941643ec3057c2/cypress/support/constants/constants.js#L1797-L1813

We can add "Save & keep editing" to the reusable object

cypress/support/fragments/orders/orderEditForm.js
```diff
@@ -54,9 +56,15 @@ const sections = {
 const buttons = {
   Cancel: cancelButton,
   'Save & close': saveAndCloseButton,
+  'Save & keep editing': saveAndKeepEditingButton,
```
usavkov-epam: I understand that the existing pattern leads to using such keys, but I suggest using square brackets notation for keys like those.

e.g.
```
{
  // ...
  [COMMON_BUTTON_LABELS.SAVE_AND_KEEP_EDITING]: saveAndKeepEditingButton,
  // ...
}
```

Single source of truth is great :)

cypress/support/fragments/orders/orderEditForm.js
```diff
@@ -26,6 +27,7 @@ const orderSummarySection = orderEditFormRoot.find(Section({ id: 'poSummary' }))
 const collapseAllButton = orderEditFormRoot.find(Button('Collapse all'));
 const cancelButton = orderEditFormRoot.find(Button('Cancel'));
 const saveAndCloseButton = orderEditFormRoot.find(Button('Save & close'));
+const saveAndKeepEditingButton = orderEditFormRoot.find(Button('Save & keep editing'));
```
usavkov-epam: Let's use constant values instead of "Magic" strings

cypress/support/fragments/orders/orderLineEditForm.js
```diff
@@ -37,6 +38,8 @@ const automaticExportInfoIconSelector = '[data-test-info-popover-trigger]';
 const cancelButton = Button('Cancel');
 const saveButton = Button('Save & close');
 const saveAndOpenOrderButton = Button('Save & open order');
+const saveAndKeepEditingButton = Button('Save & keep editing');
+const saveAndCreateAnotherButton = Button('Save & create another');
```
usavkov-epam: Also a good place to use already defined constants

cypress/support/fragments/orders/orderLineEditForm.js
```diff
@@ -78,7 +82,13 @@ const buttons = {
   Cancel: cancelButton,
   'Save & close': saveButton,
   'Save & open order': saveAndOpenOrderButton,
+  'Save & keep editing': saveAndKeepEditingButton,
+  'Save & create another': saveAndCreateAnotherButton,
```
usavkov-epam: The same place might use dynamic object keys

usavkov-epam (commented):
Thanks for the changes — overall looks good. One request: I noticed several occurrences of repeated string literals (magic strings) across the changes. Could you extract these into constants (or a small constants/enums module) and reuse them instead of repeating the literal everywhere?

Rationale:
- Single source of truth — changing the value later only requires updating one file.
- Reduces chance of typos and inconsistent values across tests/code.
- Makes intent clearer (a well-named constant is more descriptive than the raw string).

ostapwd (approved):

usavkov-epam (approved):

PR #7536, opened 2026-08-14
FAT-24304: implements C605928

zepheiryan:
https://folio-org.atlassian.net/browse/FAT-24304

Screenshot 2026-08-14 at 14 00 05


valerypilko (approved):

cypress/e2e/marc/marc-authority/edit-linked-marc-authority/update-with-data-import-check-order-of-controlled-subfields-in-bib.cy.js
```diff
@@ -0,0 +1,261 @@
...
+  authorityHeadingType: 'Geographic Name',
+  tag651: '651',
+  tag151: '151',
+  field651Index: 28,
+  field651Ind1: '\\',
+  field651Ind2: '0',
+  field651Initial: '$a United States of America $x History $e Country $b States $e USA $y Civil War, 1861-1865 $x Cavalry operations.',
+  field651Uri: 'http://id.loc.gov/authorities/subjects/sh85140220',
+  field651Uncontrolled2: '$8 number801 $1 URI1 $8 number802',
+  updatedAuthTitle: 'United States of America USA Independence war History Civil War, 1861-1865 Cavalry operations',
```
zentestuken: Please make sure that authority heading (1XX) and instance titles (245) are at least somewhat unique. The simplest approach is to add "AT_C605928" at the beginning of the 1XX $a for an authority, 245 $a for an instance. This substantially reduces the amount of data-related issues in parallel runs, runs on bugfest envs (when search/browse/import/auto-linking, etc. steps fail due to multiple matches). Also, we can clearly see that the record created by auto-tests and know exact test case ID, which may be helpful sometimes.
Please also make sure that authority naturalId is somewhat unique, f.e. by adding "605928" to 001, 010 $a in authority records.
The perfect way to ensure uniqueness would be to generate records using API methods, if possible, f.e.:
`cy.createMarcBibliographicViaAPI`
`MarcAuthorities.createMarcAuthorityViaAPI`

zentestuken (approved):

PR #7582, opened 2026-08-26
FAT-28155 Implement test cases related to "Search and filter" field of Lists app

vashjs:
Refs: [FAT-28155](https://folio-org.atlassian.net/browse/FAT-28155)


cypress/e2e/lists/verify-search-lists-case-insensitive-partial-match.cy.js
```diff
@@ -0,0 +1,145 @@
...
+        Lists.clickOnSearchButton();
+        Lists.verifyListIsPresent(createdLists[2].name);
+        Lists.verifyListIsPresent(Lists.cannedListInactivePatronsWithOpenLoans);
+        Lists.verifyListsPaneRecordsCount(2);
+
+        // #4 Search "list", click "Search" (matches anywhere in the field)
+        Lists.fillInSearchField('list');
+        Lists.clickOnSearchButton();
+        Lists.verifyListIsPresent(createdLists[3].name);
+        Lists.verifyListsPaneRecordsCount(1);
```
sviatlana-stsiapanava, line 129: most probably during parallel test run this search will return more than 1 list because the word 'list' is used in many names like this:

const listName = `AT_C451555_List_${getRandomPostfix()}`;

To be sure that the only one list will be returned I would suggest to use any other word in created list and for search, e.g. 'catalog'

cypress/e2e/lists/verify-search-lists-with-filters.cy.js
```diff
@@ -0,0 +1,234 @@
...
+      cy.getUserToken(user2.username, user2.password);
+      [list1.id, list2.id, list4.id, list5.id].forEach((id) => Lists.deleteViaApi(id));
+      cy.getAdminToken();
+      Users.deleteViaApi(user1.userId);
+      Users.deleteViaApi(user2.userId);
+    });
+
+    it(
+      'C1434639 Verify search lists used simultaneously with filters (athena)',
+      { tags: ['extendedPath', 'athena', 'C1434639'] },
```
sviatlana-stsiapanava: in TestRail this test case has 'criticalPath' test group

sviatlana-stsiapanava (approved):

PR #7584, opened 2026-08-26
UITEST-146 Update AutoSuggest interactor for axe-core upgrade.

JohnC-80:
## [UITEST-146](https://folio-org.atlassian.net/browse/UITEST-146)

For an upcoming Axe-core 4.13.0 upgrade, `axe` is more strict about which aria-attributes can be applied to certain elements. The case for Autosuggest is that the aria-expanded attribute is being applied to the parent div of the when it should only be applied to the input (as it is). Upgrading axe-core requires that the attribute be removed from that div in stripes-components, so the AutoSuggest interactor in stripes-testing needs to be updated to reflect the change:
 - it must no longer look at the parent `div` for the `open` status, but it can refer to the `input` inside of it.
 
 Related stripes-components PR: https://github.com/folio-org/stripes-components/pull/2586


zburke (approved):
❤️ 👏👏👏

